PR #5, opened 2017-02-05
Sidekiq setup added

abhinavmsra:


shared/application.rb
```diff
@@ -0,0 +1,18 @@
+require_relative 'boot'
+
+require 'rails/all'
+
+# Require the gems listed in Gemfile, including any gems
+# you've limited to :test, :development, or :production.
+Bundler.require(*Rails.groups)
+
+module Tankeboksen
```
olivierobert: @abhinavmsra what is it `Tankeboksen` here. It should come from the application name

PR #12, opened 2017-05-07
docker tips: recommended gem caching

MidnightWonderer:
I'm not sure how you guys use this template.  
But I guess you will spend a lot of time when you made a change to Gemfile.  
I'm not sure if your `cache the gem` section work properly. (Seems to me it is not)  
  
This is my recommendation about gem caching.  
Basically I think it should be cached at compose level and no `bundle install` at the docker image itself.  
Not really sure if it compatible with your current configuration though.  
  
plus some related goodies e.g. dumb-init, and up-to-date nodejs  
I personally don't use the template though.  
So no test performed.

### DIFF
--- a/rails_docker/Dockerfile
+++ b/rails_docker/Dockerfile
@@ -1,18 +1,12 @@
 FROM ruby:2.4.0
 
-RUN apt-get update -qq && apt-get install -y build-essential libpq-dev nodejs
+RUN apt-get update -qq && apt-get install -y build-essential libpq-dev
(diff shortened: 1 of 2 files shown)


olivierobert:
Thank you for your pull request and suggestion.

We have made [several changes](https://github.com/nimbl3/rails-templates/blob/master/rails_docker/docker-compose.yml) to the Dockerfile since you have opened your pull request. As rightly noted, the previous config did not use the latest caching mechanism with Docker Compose V3. This has been solved by the latest update to the repo.

We have also made [several changes](https://github.com/nimbl3/rails-templates/blob/master/rails_docker/Dockerfile) to the Dockerfile. Namely, we have added the setup of Node.JS (your script did not work on that part by the way), Yarn and PhanatomJS. We looked into `dumb-init` but we have decided not to include it for now.

PR #18, opened 2017-07-03
Hotfix - Docker issue

abhinavmsra:
Fixes:

✅ Fix db creation issue. Currently the docker app cannot connect to the db container

### DIFF
--- a/rails_docker/application.yml
+++ b/rails_docker/application.yml
@@ -4,17 +4,14 @@ default: &default
   USERNAME: "postgres"
   MAILER_DEFAULT_HOST: "localhost"
   MAILER_DEFAULT_PORT: "3000"
-  DOORKEEPER_TOKEN_PATH: "http://localhost:3000/api/oauth/token"
   MAILER_SENDER: "Test <noreply@nimbl3.co>"
 
 development:
   <<: *default
   DB_NAME: "#{app_name}_development"
-  DOORKEEPER_CLIENT_REDIRECT_URI: "urn:ietf:wg:oauth:2.0:oob"
   SECRET_KEY_BASE: "#{secret_key_base}"
 
 test:
   <<: *default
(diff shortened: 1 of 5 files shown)


rails_docker/test.sh
```diff
@@ -1,5 +1,3 @@
-#!/bin/bash
```
olivierobert: why removing the shebang line?

rails_docker/docker-compose.yml
```diff
@@ -16,6 +16,7 @@ services:
       - '3000:3000'
     volumes:
       - '.:/#{app_name}'
```
olivierobert: The volumes config need to contain the bundle volume as well to avoid rebuilding it each time:

```
volumes:
      - '.:/#{app_name}'
      - "ruby-bundle:/bundle"
```

rails_docker/docker-compose.yml
```diff
@@ -16,6 +16,7 @@ services:
       - '3000:3000'
```
olivierobert: let's order the config settings by alphabetical order + here are some improvement taken from Athena:

```
web:
  build: .
  container_name: '#{app_name}_web'
  command: bash -c "bundle check || bundle install && rm -f tmp/pids/server.pid && bundle exec rails s -p 3000 -b '0.0.0.0'"
 depends_on:
    - db
  ports:
    - "3000:3000"
  stdin_open: true
  tty: true
  volumes:
    - '.:/#{app_name}'
    - "ruby-bundle:/bundle"
```

rails_docker/application.yml
```diff
@@ -1,20 +1,17 @@
 default: &default
-  DB_HOST: "192.168.99.100"
+  DB_HOST: "db"
```
caherrerapa: shouldn't be localhost?

PR #29, opened 2017-11-17
Upgrade Rails to 5.1.4 and use System Test config in RSpec 3.7.2

anduonghien:
- [x] Upgrade Ruby to 2.4.2
- [x] Upgrade Rails to 5.1.4
- [x] Upgrade RSpec to 3.7.2
- [x] Upgrade selenium-webdriver to 3.7.0
- [x] Remove Poltergeist
- [x] Add RSpec configurations to use SystemTest from Rails
- [x] Fix warnings from Rubocop


lib/rspec.rb
```diff
@@ -9,7 +9,7 @@ def setup_rspec
   # folder for rspec
   run 'mkdir spec/codebase'
   run 'mkdir spec/fabricators'
-  run 'mkdir spec/features'
+  run 'mkdir spec/systems'
```
olivierobert: Is it `systems` or `system` in the end? I thought we discuss about it during Nimbl3 Growth and you said it was `system` 😜

rails_docker.rb
```diff
@@ -29,7 +29,7 @@ def source_paths
 # Gemfile
 remove_file 'Gemfile'
 copy_file 'rails_docker/Gemfile.txt', 'Gemfile'
-copy_file 'rails_docker/.rvmrc', '.rvmrc'
+copy_file 'rails_docker/.ruby-version', '.ruby-version'
```
olivierobert: Is this compatible with an version package manager?

shared/rspec/rails_helper.rb
```diff
@@ -11,6 +11,7 @@
 ActiveRecord::Migration.maintain_test_schema!
 
 RSpec.configure do |config|
+  # rubocop:disable Lint/InterpolationCheck
   config.fixture_path = "\#{::Rails.root}/spec/fabricators"
```
olivierobert: Why not fixing it with:

```
File.join(::Rails.root, 'spec', 'fabricators')
```

shared/rspec/support/system.rb
```diff
@@ -0,0 +1,7 @@
+RSpec.configure do |config|
+  config.before(:each, type: :system) do
+    # driven_by :selenium, using: :chrome
+    # driven_by :selenium, using: :firefox
```
olivierobert: Let's remove this commented code. Devs can also switch the driver per spec:

```
it 'creates a new record', driver: :firefox do
   ...
end
```

shared/rspec/support/retry.rb
```diff
@@ -6,8 +6,8 @@
   # show exception that triggers a retry if verbose_retry is set to true
   config.display_try_failure_messages = true
 
-  # run retry only on features tests
-  config.around :each, :js do |example|
+  # run retry only on systems tests
+  config.around :each, type: :system do |example|
     example.run_with_retry retry: 3
```
olivierobert: How about adding an env var to do the retry only when needed e.g. CI -> `TEST_ENABLE_RETRY`. 

```
example.run_with_retry retry: 3 if ENV.fetch('TEST_ENABLE_RETRY`, true)
```

Usually in local dev, we don't want to retry as we want it to fail fast. But in the CI env, to prevent false negative we usually enable it.

I think this has be done in a few projects. What do you think?

anduonghien: Agree 👍

shared/rspec/rails_helper.rb
anduonghien: @olivierobert I think this one `Rails.root.join('spec', 'fabricators')` is enough

shared/rspec/support/system.rb
anduonghien: @olivierobert Currently RSpec doesn't support to change driver by use `driver: :firefox`

If devs want to change it for specific test, dev should change like this

```
it 'creates a new record' do
  driven_by :selenium, using: :firefox
  ...
end
```

So not sure should I keep the commented code or remove it, if I keep it, devs can be copy and paste faster than write, also if they want to change the driver at global scope, just uncomment, no need to write again. 😄 

How do you think?

lib/rspec.rb
anduonghien: From Rspec, they said that it is `system` because they (we also) use `config.infer_spec_type_from_file_location!` in `rails_helpers` it mean that every file under `system` folder it will run with `system type`

But for us, I think for the convention of the template, we have `models, controllers, ...` I think it will good if we go with `systems`

And if we go with `systems`, we have to put `type: :system` on each file like this

```
describe 'Sign in', type: :system
  ...
end
```

How do you think?

rails_docker.rb
anduonghien: Yes @olivierobert This is warning on my terminal when I use `.rvmrc`

screen shot 2017-11-20 at 11 17 30 am

lib/rspec.rb
abhinavmsra: @anduonghien @olivierobert if we still differentiate the specs with `type: :system`, why not have the same old `features` directory. For me its more descriptive than `system(s)`

rails_docker.rb
abhinavmsra: @anduonghien @olivierobert In the [DOCS](https://rvm.io/workflow/projects#project-file-rvmrc),

> `.rvmrc` has one major flaw - it requires trusting to prevent execution of unauthorized code, which makes it hard to use and complicates deployment to production. **In cases where scripting is not required, use of other project files is encouraged.**

I think `.ruby-version` file is a good-fit for us.

But I suggest using the shell command,

`rvm --ruby-version use [RUBY_VERSION]@[GEMSET]`

This way we can generate the ruby-version file and the gemset file at once.

rails_docker/Gemfile.txt
```diff
@@ -52,19 +52,19 @@ group :development, :test do
...
   gem 'spring-commands-rspec' # This gem implements the rspec command for Spring.
-  gem 'spring-watcher-listen', '~> 2.0.0' # Makes Spring watch the filesystem for changes using Listen
-  gem 'listen', '~> 3.0.5' # Listens to file modifications
+  gem 'spring-watcher-listen', '2.0.1' # Makes Spring watch the filesystem for changes using Listen
+  gem 'listen', '3.1.5' # Listens to file modifications
   gem 'letter_opener' # Preview mail in the browser instead of sending.
 end
 
 group :test do
+  gem 'webmock' # Library for stubbing and setting expectations on HTTP requests in Ruby
```
abhinavmsra: @anduonghien any reason to add this, considering we already have `vcr`

shared/rspec/support/capybara.rb
```diff
@@ -1,27 +1,15 @@
...
-Capybara.register_driver :poltergeist do |app|
-  options = {
-    js_errors: false,
-    phantomjs_options: [
-      '--ssl-protocol=any'
-    ],
-    window_size: [2048, 2048]
-  }
-  Capybara::Poltergeist::Driver.new(app, options)
+# https://github.com/rails/rails/pull/30876
```
abhinavmsra: @anduonghien i like the configuration from [thoughtbot](https://robots.thoughtbot.com/headless-feature-specs-with-chrome)

Major advantage: Ease of switching between headless and browser

```
require "selenium/webdriver"

Capybara.register_driver :chrome do |app|
  Capybara::Selenium::Driver.new(app, browser: :chrome)
end

Capybara.register_driver :headless_chrome do |app|
  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
    chromeOptions: { args: %w(headless disable-gpu) }
  )

  Capybara::Selenium::Driver.new app,
    browser: :chrome,
    desired_capabilities: capabilities
end

Capybara.javascript_driver = :headless_chrome
```

shared/rspec/support/system.rb
abhinavmsra: @anduonghien its because you only register `headless_chrome ` driver in `shared/rspec/support/capybara.rb`.

We need to have multiple defined drivers in case we want to switch

shared/rspec/support/retry.rb
```diff
@@ -6,8 +6,8 @@
   # show exception that triggers a retry if verbose_retry is set to true
   config.display_try_failure_messages = true
 
-  # run retry only on features tests
-  config.around :each, :js do |example|
-    example.run_with_retry retry: 3
+  # run retry only on systems tests
+  config.around :each, type: :system do |example|
+    example.run_with_retry retry: ENV.fetch('TEST_RETRY').to_i
```
abhinavmsra: @anduonghien personally i dont think this fits to be a `ENV ` variable. It could very well have been a constant like `RETRY_COUNT` inside the same file, at max.

lib/rspec.rb
olivierobert: Let's settle on `system` as 1) it's the default name and 2) it remove the need to add `type: :system` to all tests (time saving FTW)

rails_docker/Gemfile.txt
olivierobert: @abhinavmsra `webmock` is a dependency of `vcr`. But `webmock` allows you to do things outside of `vcr` e.g. custom Mock, using `stub_request`

shared/rspec/support/retry.rb
olivierobert: @abhinavmsra I recommend An to use an env variable as it can be switched on/of without modifying any file. Think that this will be enabled on CI only

abhinavmsra: @olivierobert its something we rarely tweek

lib/rspec.rb
anduonghien: Ok @olivierobert, will change it

shared/rspec/support/system.rb
anduonghien: I'm not sure I understand what you mean @abhinavmsra 🤔

shared/rspec/support/capybara.rb
anduonghien: @abhinavmsra Acutally this configuartion is old and it only for Capybara, the Rails team had implement everything, if we go with config like that, I think it doesn't make sence because we don't use any config from System Test.

How do you think?

shared/rspec/support/system.rb
abhinavmsra: @anduonghien as we discussed, lets add some comments to explain the drivers are available by default and can be changed easily

shared/rspec/support/capybara.rb
abhinavmsra: @anduonghien @olivierobert it seems its not required any longer.

rails_docker.rb
```diff
@@ -49,6 +48,10 @@ def source_paths
 copy_file 'rails_docker/test.sh', 'bin/test.sh' # shell script for run tests on docker
 run 'chmod +x bin/test.sh'
 
+# rvm
+run 'touch .ruby-version && echo 2.4.2 > .ruby-version'
+run "touch .ruby-gemset && echo #{app_name} > .ruby-gemset"
```
olivierobert: Nice ❤️  to use gemsets

rails_docker/Gemfile.txt
```diff
@@ -20,7 +20,6 @@ gem 'rails_admin'
 
 # Authentications & Authorizations
 gem 'devise' # Authentication solution for Rails with Warden
-gem 'doorkeeper' # OAuth 2 provider
 gem 'pundit' # Minimal authorization through OO design and pure Ruby classes
 
 source 'https://rails-assets.org' do
```
olivierobert: Can we have another PR to remove this? Don't think we are going to `https://rails-assets.org` with Webpack

PR #30, opened 2017-11-20
Remove test container, remove the PhantomJS

anduonghien:


anduonghien:
@olivierobert @abhinavmsra As we discussed yesterday, we will remove the `Test container` and run the test on local, keep the web run on docker.

But me and @abhinavmsra have a discussed today. If we move the test out side, it isn't a full docker app, we have the WEB run in docker and the TEST run on local machine, so in the end the dev need to install the `ruby, ...` on local machine as well.

Also another thing is if we move the TEST to run on local machine, the connection to service (like database) is different between development and test.
Example on development (web container) we connect to DB with the host is `db` (connect docker container to docker container) but on the test, the DB host is `localhost` (connect from local machine to docker container).

If we run the test inside the docker, it's full docker app but we don't get the Chrome UI for feature test yet.

So I just raise an issue at here and we will continue disscuss Should will go with full docker app or with full local app?

abhinavmsra:
In summary,

**Cons of Dockerizing Tests**

1. Lose Chrome UI for system tests.

Probable Solution:

1. Just a thought: If we can symlink the chrome directories between local machine and docker container, **MAYBE** when the tests are executed inside container, chrome UI is triggered in the local machine.

**Cons of NOT Dockerizing Tests**

1. All the Rails setup has to be moved to local machine again, which is turn defeats the purpose of dockerizing `web` in the first place.

Probable Solution:

1. Symlink the `BUNDLE_PATH` directory so that when we run the tests, gems are still available.

2. Override the `HOST` parameters in TEST ENV.

olivierobert:
@abhinavmsra @anduonghien summary of what we discussed:

* We keep the Dockerfile as-is to be able to 1) deploy the app in production as a dockerised app and 2) to be able to run the tests in a docker CI e.g. Semaphore Docker, Codeship Docker or Codefresh
* Add a `docker-compose.dev.yml` to run the web app locally
* Keep `docker-compose.yml` to run everything as a Dockerised app

anduonghien:
@abhinavmsra Also don't forget to put the `Procfile` and `Procfile.dev` as well 💜

rails_docker.rb
```diff
@@ -77,18 +83,18 @@ def source_paths
...
-  remove_file "Guardfile"
+  remove_file 'Guardfile'
   copy_file 'shared/Guardfile', 'Guardfile'
 
   #create .rubocop.yml
   copy_file 'shared/.rubocop.yml', '.rubocop.yml'
 
   #shell script for run database on docker
-  copy_file 'rails_docker/envsetup.sh', 'envsetup.sh'
+  copy_file 'rails_docker/envsetup.sh', 'bin/envsetup.sh'
```
anduonghien: @abhinavmsra If we moved to inside the bin, we can remove the `.sh` extension.

PR #31, opened 2017-11-27
Release/5.1.4: Upgrade to Rails 5.1.4

abhinavmsra:


lib/test_env.rb
```diff
@@ -0,0 +1,13 @@
+def setup_test_env
+  _disable_animations
+end
+
+def _disable_animations
```
anduonghien: @abhinavmsra I think we don't use this convention (I mean underscore before the method name)

lib/rspec.rb
```diff
@@ -9,7 +9,7 @@ def setup_rspec
   # folder for rspec
   run 'mkdir spec/codebase'
   run 'mkdir spec/fabricators'
-  run 'mkdir spec/features'
+  run 'mkdir spec/system'
```
caherrerapa: I don't like 'systems' but would adhere to conventions e.g. models, requests etc.

olivierobert: @caherrerapa this is a new spec type introduced by Rails 5.1 so it's standard. We already had a discussion [here](https://github.com/nimbl3/rails-templates/pull/29#discussion_r151890678)

Issue #34, opened 2017-12-06
Add `gem webpacker`


Lahphim:
Should we add `gem webpacker` to the `Gemfile.txt`?
Because It has some issues when trying to create a new project with React environment using webpack.

I use this command `--webpack=react`.

olivierobert:
@Lahphim yes we could. But maybe we can add it conditionally: if the option `webpack` is passed in the rails new command then it could be added, otherwise not. How about you having a dab at it?

olivierobert:
Implemented in #39

PR #39, opened 2017-12-19
Added webpacker setup.

abhinavmsra:
#34 

**Features Added**

✅ Setup for webpack.


shared/webpack/loaders/coffee.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  test: /\.coffee(\.erb)?$/,
```
olivierobert: do we really need this as we don't use Coffescript?

abhinavmsra: @olivierobert I kept it because many libraries still use `.coffee` like RailsAdmin

olivierobert:
@abhinavmsra why closing it?

PR #40, opened 2018-01-02
Production-ready Docker-based deployment

olivierobert:
Improve the Docker configuration for tests and make it ready for Docker-based deployment.

✅ Move all test env configuration to a separate Docker Compose config `docker-compose.test.yml`
✅ Improve the main `Dockerfile` to make it ready for Docker-based deployment. Tested on Heroku.
✅ Use `ruby:2.4.2-slim` as it's much smaller that the default `ruby:2.4.2` image. So huge speed gains in tests builds and deployments 🚀
✅ Leverage Docker Compose cache settings to speed up build settings. This prevents rebuilding the whole Docker image on each build.

The configuration is broken down into the following:

rails-templates___volumes_workspace_rails-templates__-_____rails_docker_docker-compose_yml__rails-templates_

The build flow for tests is the following:

* Each branch has a tagged Docker image on Docker Hub. 

nimbl3_athena-web_-_docker_hub

This image will be used to cache built images. Currently, the tag uses the SEMAPHORE_BRANCH_ID as BRANCH_NAME cannot be used since our branches have special characters e.g. `chore/setup-docker` which is invalid as image tags.

* On each build, pull the Docker image to use it as a cache base.

* Rebuild the Docker image. Most steps are now using cache and only parts that changed will be rebuilt:

_passed___build_40_of_athena-web_chore_setup-deploy-docker-to-heroku_-_semaphore

The longest build step is currently the assets pre-compilation which cannot and should not be cached.

* Push the updated Docker image on Docker Hub

The build flow for deployment is the following:

* Push the Docker branch tagged `latest` to the Heroku Registry.

```
heroku container:push web
```

* On each update to the Heroku Registry, the application is [automatically deployed](https://devcenter.heroku.com/articles/container-registry-and-runtime#getting-started).

The following registries are setup on the CI server:

semaphore


rails_docker/Dockerfile
```diff
@@ -36,8 +37,26 @@ RUN locale-gen en_US.UTF-8
...
+COPY . ./
+
+# Compile assets
+RUN RAILS_ENV=$RAILS_ENV bundle exec rake assets:precompile
+
+# Make the init file executable
+RUN chmod +x ./bin/start.sh
 
-RUN bundle install
+EXPOSE $PORT
```
abhinavmsra: @olivierobert how about managing all the ports from docker-compose.yml file.

rails_docker/Dockerfile
```diff
@@ -1,32 +1,33 @@
-FROM ruby:2.4.2
+FROM ruby:2.4.2-slim
```
abhinavmsra: @olivierobert i think its better to split Dockerfile based on ENVs, renaming this to `DockerfileCI`.

rails_docker/docker-compose.yml
```diff
@@ -1,36 +1,37 @@
...
-    entrypoint: bin/test.sh
+      - "${PORT}:${PORT}"
     stdin_open: true
     tty: true
+    depends_on:
+      - db
     volumes:
-      - .:/#{app_name}
       - "ruby-bundle:/bundle"
     environment:
```
abhinavmsra: @olivierobert its very likely that we will have many env variables. 

Lets use [env_file](https://docs.docker.com/compose/environment-variables/#the-env_file-configuration-option) configuration option instead, so that we dont have to change `docker-compose.yml` file each time we add a new env

shared/README.md
```diff
@@ -38,12 +38,72 @@ brew install yarn
...
+* Make sure that the config file `pronto.yml` for pronto has been added as a configuration file.
+ 
 * In the CI build script, add the following command:
         
-  ```
-    if ([ $BRANCH_NAME != 'master' ] && [ $BRANCH_NAME != 'development' ]); then (echo "Running pronto"; bundle exec pronto run -f bitbucket_pr -c origin/development); else (echo "Escaping pronto"); fi
-  ```
\ No newline at end of file
+```
+if ([ $BRANCH_NAME != 'master' ] && [ $BRANCH_NAME != 'development' ]); then (echo "Running pronto"; bundle exec pronto run -f bitbucket_pr -c origin/development); else (echo "Escaping pronto"); fi
```
abhinavmsra: @olivierobert could you please add ` || true` to make it fail safe, as you discovered was the case when no Pull Request was created

rails_docker/Dockerfile
olivierobert: NOt sure about having different Dockerfile as 1) it takes load of time to test and make it work and 2) there are no differences between the `test` (`docker-compose.test.yml`) and `production` (`docker-compose.yml`) environments which are the two envs where the web app is run inside a Docker image. Using the slim version has no downsides in our case as our envs are pretty standard.

rails_docker/Dockerfile
olivierobert: It's is already in `docker-compose.yml`. But I could add it for `docker-compose.test.yml` too. But the thing is that it would still be needed in the Dockerfile as not all providers have Docker Compose deployment e.g. Heroku only has Docker ddeployments, no support for Docker Compose. So it's safer to still have it in the Dockerfile.

rails_docker/docker-compose.yml
olivierobert: I added the env file `.env` which is loaded automatically but then we would need to split it be environments e.g. `.env.test` and `.env.production`. My main motivation not to do it is that it's not supported in Heroku and it would be better to use docker-override yml file.

PR #43, opened 2018-01-10
Disable instance variable assumption detector for controllers

rosle:
In Braive, we have many usage of the pattern where we set the instance through `before_action` like this https://github.com/troessner/reek/issues/1133, which the error should not be raised. 

And also in controllers, there might be many used of the instance variables. It might be good if we disable it.


olivierobert:
@rosle how we add all the recommended settings for Rails: https://github.com/troessner/reek/blob/master/README.md#working-with-rails

PR #45, opened 2018-01-30
Update transactional fixtures for using in system test

Lahphim:
We should set it to `true` because of use in System test.


shared/rspec/rails_helper.rb
```diff
@@ -13,7 +13,7 @@
 RSpec.configure do |config|
   config.fixture_path = Rails.root.join('spec', 'fabricators')
 
-  config.use_transactional_fixtures = false
+  config.use_transactional_fixtures = true
```
anduonghien: Let add the comment here @Lahphim 👍

Lahphim: @anduonghien Added

PR #53, opened 2018-05-07
System Specs config improvement

olivierobert:
## What happened
The work started due to Chrome driver stability issues (in #51). Upon research and fixing the issues, I found a more complete and also more transparent way to configure Capybara for our projects.

This solution uses the new flag `disable-dev-shm-usage` available since Chrome 65 (see [info](https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/docs/troubleshooting.md#tips)). Contrary to #51, it does not require to upgrade to Docker Compose 3.5 ot latest which helps in supporting more CI/CD providers.
 
## Insight
* Supplanting work done in #51 
* Inspired by https://gitlab.com/gitlab-org/gitlab-ce/blob/master/features/support/capybara.rb
* Goodie: execute a test in non-headless Chrome on your local machine by exporting this env variable: `export CHROME_HEADLESS=false`. Then reset it: `unset CHROME_HEADLESS`.


shared/rspec/support/capybara.rb
```diff
@@ -1,26 +1,46 @@
 require 'selenium-webdriver'
 require 'capybara'
+require 'capybara/rspec'
+
+# Give CI some extra time
+CAPYBARA_TIMEOUT = ENV['TEST_CI_SERVER'] ? 60 : 30
```
rosle: While checking the CI services, I notice that most of it have `CI=true` variable set. Not sure, should we use that variable for checking or not?

olivierobert: Great idea. Will change to this variable 👍

olivierobert:
@rosle changed in 152a7a9

PR #54, opened 2018-05-31
CI Improvements

rosle:
## What happened
✅ Moving the configuration into the project, So we can easily setup the Semaphore for the new project.
✅ Adding the screenshot upload command to upload the screenshot to S3 bucket.

 
## Insight
- Add the `.semaphore.yml` with all the configurations that we can copy into the project. For the config, I basically took the current config from the docker-based project on Semaphore.
- For the screenshots, it will upload all the screenshots in `tmp/screenshots` to s3 bucket configured in the setting. After completed, it will print out the accessible url, so we can check the failed test more easily.
- Because we use Docker, the screenshots volume should be mounted, so we can access the screenshot files in the container. 

## Proof Of Work
screen shot 2561-05-31 at 12 26 08

screen shot 2561-05-31 at 12 26 27


shared/.semaphore.yml
```diff
@@ -0,0 +1,38 @@
+# Setting for Semaphore CI
+# Copy these configurations to the settings on Semaphore CI
+
+project_settings:
+  env_variables:
+    NODE_ENV: production
```
abhinavmsra: @rosle Setting `NODE_ENV` to production will force `webpacker` to build using production config. 
Should be `test` IMO.

https://github.com/rails/webpacker#custom-rails-environments

abhinavmsra: Also `envs` are defined in `docker-compose` file. Would be better to have all the env configs at one place.

```
- RACK_ENV=test
  RAILS_ENV=test
```

rosle: @abhinavmsra Not sure about this, I just grab this the config we set in the semaphore project, I'll need to recheck it again. 👍

olivierobert: @rosle I think @abhinavmsra is right. Let's change it in this PR and in the projects on Semaphore 🙇

Issue #60, opened 2018-08-27
Use Sass functions for convert `px` unit in stylesheets.


Yutna:
Convert CSS unit from `px` to `rem` or `em` is a very tedious task. For improve more productive when you are working on stylesheets. `rails-templates` will provide `rem` and `em` Sass functions.

These functions will look like this when we are using.

**`rem()`**
```scss
.sample-1 {
  padding: rem(20px); // padding: 1.25rem;
}

.sample-2 {
  padding: rem(20px 16px); // padding: 1.25rem 1rem;
}

.sample-3 {
  padding: rem(20px 16px 32px); // padding: 1.25rem 1rem 2rem;
}

.sample-4 {
  padding: rem(20px 16px 32px auto); // padding: 1.25rem 1rem 2rem auto;
}
```

**`em()`**
```scss
.sample-1 {
  padding: em(20px); // padding: 1.25em;
}

.sample-2 {
  padding: em(20px 16px); // padding: 1.25em 1em;
}

.sample-3 {
  padding: em(20px 16px 32px); // padding: 1.25em 1em 2em;
}

.sample-4 {
  padding: em(20px 16px 32px auto); // padding: 1.25em 1em 2em auto;
}
```

olivierobert:
Implemented in #61

PR #61, opened 2018-08-27
Create Sass functions for convert `px` unit in stylesheets.

Yutna:
ISSUE: #60 

## What happened
For improve more productive when you are working on stylesheets. `rails-templates` should provide `rem` and `em` Sass functions for convert `px` unit.

## Insight
The `rem` and `em` functions can accepts the list of `px` unit values. It will convert all of those values to the same unit  as the function name.

## Proof Of Work
For convenient to proof of work. You can copy the source code of `rem` and `em` function to [Sassmeister](https://www.sassmeister.com/) and try to convert `px` to `rem` or `em` unit.

**NOTE** See #60 for examples


shared/app/assets/stylesheets/functions/_rem.scss
```diff
@@ -0,0 +1,14 @@
+@function rem($values, $base-font-size: 16px) {
```
olivierobert: How about merging these two functions into one file e.g. `sizing.scss`?

PR #75, opened 2018-11-10
Upgrading Rails

### LINKED ISSUE #65: Rails 5.1.4 NOT support PG gem on version ~> 1.0.0
We can NOT use the Rails generator to create the model since the Rails version 5.1.4 NOT support `pg` gem on version `1.0.0`.

We need to set the `pg` gem in Gemfile to something like this.

```ruby
gem 'pg', '~> 0.21.0'
```

caherrerapa:
## What happened
Upgrading to rails 5.1.6

 
## Insight
We need to upgrade as 5.1.4 doesn't support pg > 1.0
 

## Proof Of Work
Simple change
Fixes #65

### DIFF
--- a/rails_docker/Gemfile.txt
+++ b/rails_docker/Gemfile.txt
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 ruby '2.4.2'
 
 # Backend
-gem 'rails', '5.1.4' # Latest stable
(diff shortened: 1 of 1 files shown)


olivierobert:
This is a fine intermediary step but we should actually move this to 5.12 raise in #56

PR #83, opened 2019-01-09
Fix missing Redis configuration on Docker compose file. 

### LINKED ISSUE #67: Missing Redis configuration in Docker file.
I can NOT run the `Procfile.dev` to start the application server because it missing the Redis configuration in the docker file.

please add this to the docker file.

```yml
redis:
    image: redis:4.0.9
    container_name: <app-name>_redis
    ports:
      - "6379:6379"
```

Yutna:
Fixes #67 

## What happened
 As a description in #67, We can NOT run the `Procfile.dev` file because we are missing the Redis configuration in the Docker compose.
 
## Insight
Just as Redis configuration to the Docker compose file. 

## Proof Of Work
You should start the Redis with `Procfile.dev` file

### DIFF
--- a/rails_docker/docker-compose.yml
+++ b/rails_docker/docker-compose.yml
@@ -9,6 +9,12 @@ services:
     ports:
       - "5432:5432"
 
+  redis:
+    image: redis:4.0.9
+    container_name: #{app_name}_redis
+    ports:
+      - "6379:6379"
+
   web:
     build:
       context: .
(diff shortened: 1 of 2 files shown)


rails_docker/docker-compose.yml
```diff
@@ -9,6 +9,12 @@ services:
     ports:
       - "5432:5432"
 
+  redis:
```
anduonghien: @Yutna I think we don't need this for Production, normally for Production we will use another Service for Redis like https://elements.heroku.com/addons/redistogo

olivierobert: Let's leave it by default so that it can be removed when needed

PR #88, opened 2019-04-10
Update Rails from 5.1.6 to 5.2.3

rosle:
## What happened

✅ Upgrade Rails to 5.2.3
 
## Insight

 Upgrading rails and add some libraries that is missing + new from rails 5.2

New gems: 💎 

- bootsnap
- chromedriver-helper

New configuration files: 📁 

- /config/credentials.yml.enc (A replacement for secret.yml)
- /config/master.key
- /config/initializers/content_security_policy.rb
- /config/storage.yml

## Proof Of Work

1. Create the project from the template

2. Start the rails app with
```
bin/envsetup
foreman start -f Procfile.dev
```

Screen Shot 2562-04-10 at 12 31 22

### DIFF
--- a/rails_docker/Gemfile.txt
+++ b/rails_docker/Gemfile.txt
@@ -49,7 +49,6 @@ group :development, :test do
   gem 'shoulda-matchers' # Tests common Rails functionalities
   gem 'capybara', '>= 2.15' # Integration testing
   gem 'selenium-webdriver', '3.7.0' # Ruby bindings for Selenium/WebDriver
-  gem 'chromedriver-helper' # Easy installation and use of chromedriver to run system tests with Chrome
   gem 'database_cleaner' # Use Database Cleaner
 
   gem 'pry-rails' # Call 'binding.pry' anywhere in the code to stop execution and get a debugger console
--- a/rails_docker/Gemfile.txt
+++ b/rails_docker/Gemfile.txt
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 ruby '2.5.3'
 
 # Backend
-gem 'rails', '5.1.6' # Latest stable
+gem 'rails', '5.2.3' # Latest stable
 gem 'pg' # Use Postgresql as database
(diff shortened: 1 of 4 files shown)


rails_docker/Gemfile.txt
```diff
@@ -45,8 +47,9 @@ group :development, :test do
   gem 'rspec-rails' # Rails testing engine
   gem 'guard-rspec' # Auto-run specs
   gem 'shoulda-matchers' # Tests common Rails functionalities
-  gem 'capybara' # Integration testing
+  gem 'capybara', '>= 2.15' # Integration testing
   gem 'selenium-webdriver', '3.7.0' # Ruby bindings for Selenium/WebDriver
+  gem 'chromedriver-helper' # Easy installation and use of chromedriver to run system tests with Chrome
```
anduonghien: We don't need this gem @rosle, we will use the Chrome driver that was installed in the Dockerfile

olivierobert: @anduonghien @rosle I will create another issue to clean up the default gems in the Gemfile as there are lots we don't need anymore. SO this can be done here in another PR.

anduonghien: Yeah, but we can remove this gem from here too because it's a new gem, and to clean up the Gemfile, we will do it in another PR

olivierobert: Agreed. Let's do it here. New issue for the other gems: https://github.com/nimblehq/rails-templates/issues/89

rosle: @anduonghien @olivierobert But on the development now, we're not running with docker, correct?. But I think I can remove it for now because currently we install it locally, right?. 

I will remove it in this pr. We can revisit the setup later in another pr 👍

anduonghien: On development, I think we don't need the Google chrome driver 🤔 we need it only for test, I think

Issue #20, opened 2017-07-12
Change default configs in test env


abhinavmsra:
1. Enable `eager_load` of classes.

2. Enable `Bullet` errors to indicate N+1 queries.

rosle:
@abhinavmsra I created a pr for enabling the Bullet 👍. But for enabling `eager_load` on test. I am not sure. What is the reason behind it?

anduonghien:
This is in the default **test.rb** file generated by Rails, I think we could keep it like that as default, we could turn it on in the specific project if needed @abhinavmsra @rosle 

![image](https://user-images.githubusercontent.com/11751745/90375707-868c8780-e09f-11ea-8fae-0af5599f297e.png)

andyduong1920:
Closing as the template has **bundle exec rake zeitwerk:check** in [codebase_spec.rb](https://github.com/nimblehq/rails-templates/blob/develop/spec/codebase/codebase_spec.rb#L13) and it covers the eager load issue already

PR #98, opened 2019-04-17
Enable Bullet to watch for N+1 queries

rosle:
## What happened

✅ Add Bullet Configurations for `development` and `test`
 
## Insight

Bullet gem is installed but it is not configured. (Issue #20 - Change default configs in test env)

Here I added configurations for development and test. I get the configurations from vcliq project.

### Development

The warning is shown at the bottom of the screen

Screen Shot 2562-04-17 at 11 52 12

#### Test

The test will failed if there is n+1 queries

Screen Shot 2562-04-17 at 12 29 47

After fixing the issue, the test is passed

Screen Shot 2562-04-17 at 12 31 37

### DIFF
--- a/lib/bullet.rb
+++ b/lib/bullet.rb
@@ -1,0 +1,28 @@
+def setup_bullet
+  environment(nil, env: 'development') do
+    <<~EOT
+      # Configure Bullet gem to detect N+1 queries
+      config.after_initialize do
+        Bullet.enable        = true
+        Bullet.bullet_logger = true
+        Bullet.console       = true
+        Bullet.rails_logger  = true
+        Bullet.add_footer    = true
+      end
+
+    EOT
+  end
+
+  environment(nil, env: 'test') do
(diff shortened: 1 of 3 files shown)


anduonghien:
@rosle I'm not sure we make the test is failed if there are any N+1 queries, in somecase, we can't avoid the N+1, I remember for the Questionnaire in Braive, we have N+1 there but it hard to fix that...

Hmm but maybe it good to add that to the Test in the template, and then depend on the project, we can disable it. What do you think?

rosle:
@anduonghien Yes I think it's good if we enable it as default. And we can use this -> https://github.com/flyerhzm/bullet#whitelist to disable it in some place because I think it is a rare case. 🤔 Do you agree?

olivierobert:
@rosle @anduonghien I am for making it by default then exclude it if not needed. This would be similar to Rubocop.

PR #99, opened 2019-04-17
Replace chromedriver-helper with webdrivers

rosle:
## What happened

✅ Replace `chromedriver-helper` with `webdrivers`

## Insight

Now `chromedriver-helper` is deprecated and it is moved to `webdrivers`. On new Rails 6, it has been moved to `webdrivers` too. See https://github.com/rails/rails/pull/35732

I know we haven't include the chromedriver-helper in our gemfile before. But maybe we should? So that we don't need to install `chromedriver` on each machine that we've been working on.

## Proof Of Work

The system test is still runnable.

Screen Shot 2562-04-17 at 17 35 02

To update, we can set the version in `support/webdrivers.rb`
```
Webdrivers::Chromedriver.version = '2.46'
```

The webdriver is installed at `~/.webdriver` by default. After update the version in the config and run the spec again, it should update the installed binary.

Screen Shot 2562-04-17 at 18 16 37

### DIFF
--- a/rails_docker/Dockerfile
+++ b/rails_docker/Dockerfile
@@ -21,9 +21,6 @@ ENV BUNDLE_GEMFILE=$APP_HOME/Gemfile \
 
 ENV NODE_VERSION="8"
 
-ENV CHROMEDRIVER_VERSION="2.39" \
-    CHROMEDRIVER_DIR=/chromedriver
-
 ENV LANG="en_US.UTF-8" \
     LC_ALL="en_US.UTF-8" \
     LANGUAGE="en_US:en"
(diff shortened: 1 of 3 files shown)


olivierobert:
Great idea @rosle 

My only concern is with "The drivers will now be automatically downloaded or updated when you launch a browser through Selenium.". Especially on CI platform, would that mean that in each task it will download the latest version 🤔? Probably [pinning](https://github.com/titusfortner/webdrivers#version-pinning) version could be good. What do you think?

rosle:
@olivierobert For the versioning, we can pin the version. Right now I set it in the `spec/support/webdrivers.rb`

```
Webdrivers::Chromedriver.version = '2.46'
```

If we don't specify this, it will download the latest one 👍

PR #106, opened 2019-04-29
Use ruby alpine as the image base

### LINKED ISSUE #58: Use ruby alpine as the image base
In order to reduce the size of the Docker image, we should use the alpine version for Ruby. `ruby:2.4.2-slim` is slim but a project image still weighs at around 400MB or so.

Link to available version: https://hub.docker.com/_/ruby/

svnlto:
fixes #58 

## What happened
Updated Dockerfile to source from `FROM ruby:2.6.3-alpine3.9`


rails_docker/Dockerfile
```diff
@@ -12,58 +12,63 @@ ENV RACK_ENV=$RUBY_ENV \
     BUILD_ENV=$BUILD_ENV \
     ASSET_HOST=$ASSET_HOST
 
-ENV APP_HOME=/#{app_name} \
-    PORT=80
+ENV APP_HOME=/usr/src/#{app_name}
```
olivierobert: Any benefits to move to this subdirectory?

rails_docker/Dockerfile
```diff
@@ -12,58 +12,63 @@ ENV RACK_ENV=$RUBY_ENV \
...
+        imagemagick \
+        ruby \
+        ruby-bundler \
+        ruby-irb \
+        libxml2-dev \
+        libxslt-dev \
+        libffi-dev \
+        zlib-dev \
+        git \
+        postgresql \
```
olivierobert: postgresql is generally set up as a separate service hence does not need be set up in the web app container

svnlto: Ah yes, of course. Will remove.

rails_docker/Dockerfile
svnlto: Just a convention. Would you prefer to have it in the root?

svnlto:
> Has it been tested with a fresh installation of a Rails app? If yes, it would be good to add this kind of info in the PR description.

Just a once over, "does it build without crashing". Will come up with a test case.

rails_docker/Dockerfile
svnlto: turns out we need the `postgresql-dev` package to be able to install the `pg` gem.

lib/i18n.rb
```diff
@@ -1,6 +1,6 @@
 def setup_rails_i18n
   inject_into_class 'app/controllers/application_controller.rb', 'ApplicationController' do
-    "\tinclude Localization\n"
+    "  include Localization\n"
```
olivierobert: Why the change here 🤔 Was there any issue with `\t`? Also we keep `\n` so it seems rather inconsistent

lib/i18n.rb
```diff
@@ -19,8 +19,8 @@ def setup_i18n_js
 
   insert_into_file 'app/views/layouts/application.html.erb', after: /<%= stylesheet_link_tag.*\n/ do
     <<~EOT
-      \t\t<%= javascript_include_tag 'i18n' %>
-      \t\t<%= javascript_include_tag 'translations/translations' %>
+          <%= javascript_include_tag 'i18n' %>
+          <%= javascript_include_tag 'translations/translations' %>
```
olivierobert: Same comment here as the previous about the usage of lien break `\t`

rails_docker.rb
```diff
@@ -1,25 +1,28 @@
...
-        File.expand_path(File.dirname(__FILE__))
-      end
+    if __FILE__ =~ %r{\Ahttps?://}
+      require 'tmpdir'
+      tempdir = Dir.mktmpdir("rails-templates")
+      at_exit { FileUtils.remove_entry(tempdir) }
+      git clone: [
+              '--quiet',
+              '--depth=1',
+              'https://github.com/nimbl3/rails-templates.git',
```
olivierobert: Let's fix the Github organization name to the current one:

```suggestion
              'https://github.com/nimblehq/rails-templates.git',
```

rails_docker/Dockerfile
```diff
@@ -12,78 +12,95 @@ ENV RACK_ENV=$RUBY_ENV \
...
 # RUN mkdir "$APP_HOME" "$APP_HOME/engines"
 WORKDIR $APP_HOME
 
-# Only copy the dependency definition files (Gemfile and packages) to use Docker cache for these steps
-# Install Ruby dependencies
-COPY Gemfile* ./
-# Copy each engine lib/, Gemfile and .gemspec files
-# Example:
-# COPY engines/app_auth/lib/ ./engines/app_auth/lib/
-# COPY engines/app_auth/Gemfile engines/app_auth/app_auth.gemspec ./engines/app_auth/
```
olivierobert: We need to retain the comment about copying the engines files as all apps we build use engines

rails_docker/Gemfile.txt
```diff
@@ -3,7 +3,7 @@ ruby '2.6.2'
 
 # Backend
 gem 'rails', '5.2.3' # Latest stable
-gem 'pg' # Use Postgresql as database
+gem 'pg', '1.1.4' # Use Postgresql as database
```
olivierobert: Is there a strong reason to fix the database version? If we were to start a new project, using the latest pg gem and database is preferred.

rails_docker/Gemfile.txt
```diff
@@ -73,6 +73,7 @@ group :test do
   gem 'simplecov', require: false # code coverage analysis tool for Ruby
   gem 'vcr' # Gem for recording test suite's HTTP interactions
   gem 'timecop' # Gem for time travel
+  gem 'public_suffix'
```
olivierobert: What is the gem for? As you can see all gems have a short description to document what it's used for. So you always need to add such comment. In addition, if it's not related to the task at hand, prefer opening a separate PR for this change.

lib/i18n.rb
svnlto: @olivierobert I changed those because according to `rubocop` tabs are invalid. Hence the spaces.

lib/i18n.rb
svnlto: Same as above.

rails_docker/Gemfile.txt
svnlto: Will retry without the version parameter.

rails_docker/Gemfile.txt
svnlto: will remove, not necessary anymore.

rails_docker/Dockerfile
svnlto: done

lib/i18n.rb
olivierobert: Noted

PR #108, opened 2019-05-02
Setup a test pipeline to test every pull request 

### LINKED ISSUE #76: Setup a test pipeline to test every pull request
## Rationale

As we do use the template on a regular basis (basically only when we start a new project), some issues due to outdated configuration or just plain human mistakes can make their way into the template. It's also hard to do a major Rails version jump by just changing the configuration (that's why the template is still on Rails 5.1.x 😢 ).

The solution would be to set up a test pipeline just like for any project. 

## Solution

The tests must do the following:

* Create a new Rails application based on the template. Since #70, this would not even require to clone this repo to work
* Tests that the web app can be accessed via the Docker container in all environment. Basically testing the accessibility of port 3000 (development) and 80
* Tests that a PG and Redis containers exist and can be accessed

svnlto:
Fixes #76 

## What happened

This introduces basic semaphore setup

## Insight

I've added a `makefile` that runs through the steps involved testing the template.

### DIFF
--- a/shared/README.md
+++ b/shared/README.md
@@ -2,7 +2,7 @@
 
 ## Introduction
 
-App introduction goes here ...
+> *App introduction goes here ...*
 
 ## Project Setup
 
@@ -12,8 +12,8 @@ App introduction goes here ...
 
 * Setup and boot the Docker containers:
 
-```
-./bin/envsetup
+```sh
+./bin/envsetup.sh
 ```
 
 ### Development
@@ -22,13 +22,13 @@ App introduction goes here ...
 
     * Postgres:
 
-    ```
+    ```sh
     rake db:setup
     ```
 
 * Run the Rails app
 
-```
+```sh
 foreman start -f Procfile.dev
 ```
 
@@ -40,7 +40,7 @@ Add the following build settings to run the tests in the Docker environment via 
 
 * Configure the environment variable `BRANCH_TAG` to tag Docker images per branch:
 
-```
+```sh
 export BRANCH_TAG=$SEMAPHORE_BRANCH_ID
 ```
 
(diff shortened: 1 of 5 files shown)


olivierobert:
@svnlto with which account did you set up Semaphore? I am guessing your personal account. If that's the case, this is not ideal as we will require others to have access to it for future maintenance or even to access the build logs. That's why we have a shared Github account: https://vault.nimblehq.co:8200/ui/vault/secrets/nimble/show/general/Github

Note that I changed the destination branch so that we can review only the changes for the CI setup.

shared/README.md
```diff
@@ -12,8 +12,8 @@ App introduction goes here ...
 
 * Setup and boot the Docker containers:
 
-```
-./bin/envsetup
+```sh
+./bin/envsetup.sh
```
olivierobert: The current agreed consensus is to omit the file extension for executable files. Any string reason to change that?

svnlto: 1. I didn't know about the consensus. Is this documented?
2. Mainly for consistencies sake. all other scripts in the project have the `.sh` suffix.

svnlto:
> https://vault.nimblehq.co:8200/ui/vault/secrets/nimble/show/general/Github

I can't open vault anymore for some strange reason.

shared/README.md
olivierobert: Let's go with the consistency here then let's add the extension `.sh`.

As for the consensus, it's not documented but it's just part of following what rails does as all built-in-Rail files in the bin folder do not have extensions:

mia-web___Volumes_Workspace_mia-web__-_____app_assets_stylesheets__variables_scss__mia-web_

But for our files, some had while some did not e.g. `envsetup` vs, `start.sh`

This can be fixed in another PR

svnlto: gotcha.

PR #113, opened 2019-08-30
Add package.json template and setup git hook on commit

hanam1ni:
## What happened
- Add `package.json` template and copy to generated project.
- Add default package:
  `eslint`: Javascript linter.
  `@nimbl3/eslint-config-nimbl3`: Our javascript linter configuration.
  `husky`: Setup git hook and make it shareable between team member.
- Setup hook for check commit message format to match
```
[Story ID] Commit message
```

## Proof Of Work
- `package.json` is properly generated.
rails-templates____rails-templates__-___blog_package_json

- Git hook is working. It not allow committing wrong format commit message 
blog_—_micky_Mickys-MacBook-Pro_—___blog_—_-zsh_—_238×53

### DIFF
--- a/rails_docker/.hooks/commit-msg.sh
+++ b/rails_docker/.hooks/commit-msg.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+# Get commit message
+commit_message=`cat $1`
+if [[ $commit_message =~ ^[[].{1,}[]].{1,}$ ]]
+# Commit message format is matched
+then
+  exit 0
+# Commit message format is NOT matched
+else
+  echo "Commit message format must match: \"[<User Story ID>] Commit message\""
+  exit 1
+fi
--- a/rails_docker/package.json
+++ b/rails_docker/package.json
@@ -1,0 +1,18 @@
+{
+  "name": "#{app_name}",
+  "scripts": {
+    "lint": "eslint . --color",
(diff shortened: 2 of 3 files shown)


rails_docker/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "#{app_name}",
```
olivierobert: Since all project we are for client, we should add `private": true,`

rails_docker/.hooks/commit-msg.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
olivierobert: Not sure about the file naming. `commit-message-format.sh` would be more informative

hanam1ni: @olivierobert I follow the naming from the hook name https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks (commit-msg). Because we can add more action in this hook too.

olivierobert:
I came across [Lefthook](https://github.com/Arkweid/lefthook) which could be easier to integrate (via a yml file so no bash script) and is more agnostic (works for JS and Ruby) so could be used for any kind of project. 

What do you think about replacing Husky by Lefthook.

Here is more info: https://evilmartians.com/chronicles/lefthook-knock-your-teams-code-back-into-shape

PR #122, opened 2019-10-09
Add the @nimbl3/eslint-config-nimbl3 to package.json file

anduonghien:
## What happened
Add **@nimbl3/eslint-config-nimbl3** to **package.json**

 
## Insight
We have ESlinter configuration in **shared/linters** but we didn't add the **@nimbl3/eslint-config-nimbl3** to **package.json**
 

## Proof Of Work

1/ package.json
Screen Shot 2019-10-09 at 16 18 14

2/ **nimble-eslint** is installed automatically
Screen Shot 2019-10-09 at 16 17 08

### DIFF
--- a/rails_docker.rb
+++ b/rails_docker.rb
@@ -90,10 +90,10 @@ gsub_file('package.json', %r{"turbolinks": .+\n}, '')
 insert_into_file 'app/javascript/packs/application.js', after: "require\(\"channels\"\)\n" do
   <<~EOT
 
-    import 'translations/translations'
+    import 'translations/translations';
     
-    import 'initializers/'
-    import 'screens/'
+    import 'initializers/';
+    import 'screens/';
   EOT
 end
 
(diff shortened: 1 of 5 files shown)


rails_docker.rb
```diff
@@ -104,6 +104,15 @@ def source_paths
   EOT
 end
 
+# Add @nimbl3/eslint-config-nimbl3 plugin to package.json
+insert_into_file 'package.json', before: %r{"version": .+\n} do
+  <<~EOT
+      "devDependencies": {
+        "@nimbl3/eslint-config-nimbl3": "^2.1.1"
```
olivierobert: Use a fixed version `2.1.1`.

Actually it would be good to add a dot file `.npmrc` with the content:

```
save-exact=true
```

This ensures that only fixed versions are added.

rails_docker.rb
```diff
@@ -104,6 +104,15 @@ def source_paths
   EOT
 end
 
+# Add @nimbl3/eslint-config-nimbl3 plugin to package.json
+insert_into_file 'package.json', before: %r{"version": .+\n} do
+  <<~EOT
+      "devDependencies": {
+        "@nimbl3/eslint-config-nimbl3": "2.1.1"
```
anduonghien: @olivierobert Using the fixed version

shared/.npmrc
```diff
@@ -0,0 +1 @@
+save-exact=true
```
anduonghien: Added **.npmrc** file

PR #133, opened 2019-11-10
Improve the localization concern, moving to around_action

### LINKED ISSUE #127: Using around_action to set the locale
### Why

To set the locale in `before_action`, there will is a case that the I18n can't get the default locale for some request, I faced that in a project.

For more details:

- https://github.com/ruby-i18n/i18n/issues/381
- https://github.com/rails/rails/issues/34043
- https://github.com/rails/rails/pull/34356
- https://guides.rubyonrails.org/i18n.html#managing-the-locale-across-requests

### Solution

I think we should go with the Rails way, using `around_action` instead - https://guides.rubyonrails.org/i18n.html#managing-the-locale-across-requests

andyduong1920:
Closes #127 

## Insight
Switch the locale by using **before_action** isn't safe, the locale doesn't set return correctly on some request randomly (I faced with that issue on a project).

Here is a similar issue: https://github.com/ruby-i18n/i18n/issues/381

And the official Rails guide also prefer to using around_action to switch the locale  https://guides.rubyonrails.org/i18n.html#managing-the-locale-across-requests (https://github.com/rails/rails/pull/34356)

 I also add the configuration for the I18n as since from Rails I18n 1.1.0, the config `config.i18n.fallbacks = true` (was generated by Rails) isn't correct anymore https://github.com/ruby-i18n/i18n/releases/tag/v1.1.0

## Proof Of Work

image


olivierobert:
@anduonghien rebase required

anduonghien:
Rebase @olivierobert

shared/app/controllers/concerns/localization.rb
```diff
@@ -2,12 +2,18 @@ module Localization
...
-    def set_locale
-      I18n.locale = params[:locale] || I18n.default_locale
+    def switch_locale(&action)
+      locale = extract_locale_from_param || I18n.default_locale
+
+      I18n.with_locale(locale, &action)
+    end
+
+    def extract_locale_from_param
+      I18n.available_locales.include?(params[:locale]&.to_sym) ? params[:locale].to_sym : nil
```
Tolsee: Not a big thing, but can we use `I18n.locale_available?` instead.

anduonghien: Good to know that @Tolsee 👍 Will change it

anduonghien: Changed that @Tolsee 👍

shared/app/controllers/concerns/localization.rb
```diff
@@ -2,12 +2,18 @@ module Localization
...
-    def set_locale
-      I18n.locale = params[:locale] || I18n.default_locale
+    def switch_locale(&action)
+      locale = extract_locale_from_param || I18n.default_locale
+
+      I18n.with_locale(locale, &action)
+    end
+
+    def extract_locale_from_param
+      I18n.locale_available?(params[:locale]) ? params[:locale].to_sym : nil
```
hoangmirs: Another way: `params[:locale]&.to_sym.presence_in(I18n.available_locales)`

andyduong1920:
Open on another PR https://github.com/nimblehq/rails-templates/pull/241

PR #134, opened 2019-11-10
Do not fallback to the Sprockets server when the precompiled asset is missing on Test env

anduonghien:
## What happened
Do not fallback to the Sprockets server when the precompiled asset is missing on test ENV

 
## Insight
When we use the `image_tag` sometimes we forgot to put the `image extension` like `= image_tag 'logo'` (the correct one should be `= image_tag 'logo.png'`). With that, on the Development and Test ENV everything works (because it fallback to the Sprockets server to get the assets), but when we deploy to Production, it returns a 500 error, technically it shouldn't be like that, it should fail on Test ENV first.

https://stackoverflow.com/questions/29409959/rails-assets-not-precompiling-in-production

image

## Proof Of Work

Screen Shot 2019-11-10 at 21 54 14

### DIFF
--- a/rails_docker.rb
+++ b/rails_docker.rb
@@ -209,4 +209,13 @@ after_bundle do
       environment.config.set('plugins', plugins);
     EOT
   end
+
+  # Do not fallback to assets pipeline if a precompiled asset is missed.
+  environment(nil, env: 'test') do
+    <<~EOT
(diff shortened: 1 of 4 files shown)


olivierobert:
@anduonghien rebase required

PR #143, opened 2019-12-09
Add sanitize helper in the spec

junan:
## What happened
Added `sanitize` helper method in the spec so that we can write matcher like
```ruby
expect(page).to have_content sanitize 18n.t('campaign.description_html')
``` 

 
## Insight
In the system test,  sometimes we need to match content with translated text like
```ruby
 expect(page).to have_content I18n.t('campaign.description')
```

But if the translation text is a html like `I18n.t('campaign.description_html')`
we have to do `sanitize` it like

```ruby
expect(page).to have_content  ActionView::Base.full_sanitizer.sanitize I18n.t('campaign.description_html ')
```

Which gets longer the line and not very readable

If we define the `senitize` method in the spec helper like
```ruby
def sanitize(html)
  ActionView::Base.full_sanitizer.sanitize(html)
end

```

We can test it like

```ruby
expect(page).to have_content sanitize 18n.t('campaign.description_html')
```

Which is more cleaner.

 

## Proof Of Work
N/A


anduonghien:
@junan Could you please rebase this PR 👍

shared/rspec/support/sanitize.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
```
anduonghien, line 1: the template structure was changed @junan :smile:

There is no `shared` folder anymore

image

spec/support/sanitize.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+def sanitize(html)
```
anduonghien, line 3: 1/ should we add a test for this file @junan ?

```
it 'creates sanitize file' do
    expect(file('sanitize.rb')).to exist
  end
```

2/ Should we consider apply this file for only web variants?

PR #150, opened 2020-02-10
Fix the template test

rosle:
## What happened

✅ Fix the template step that raises an error
✅ Update the testing steps in the semaphore
✅ Update the codebase to follow the Rubocop rules
✅ Add missing configurations
 
## Insight

Now the spec is failing all the time and we cannot know if the changes made is breaking the template. So here we fix the failing part to make it green again 🌲 💚 

#### Problem

1. There are errors from the template itself. It raises an error when it cannot generate the file. 
    
    Screen Shot 2563-02-10 at 15 07 15

2. The environment where it builds the app needs to be set. I found that on Semaphore, the test fails because of the Node version is too old. (Webpacker requires v 8.16+)
    
    Screen Shot 2563-02-10 at 15 06 51

   So I did specify the version of Rails and Node on the Semaphore.

And the rest is small code updates to make it passed on Rubocop.

## Proof Of Work

The test should pass now 👍


rosle:
@olivierobert Could you recheck my last commit. I just fix the failing rubocop - https://semaphoreci.com/nimble/rails-templates/branches/chore-fix-testing-pipeline/builds/17

PR #155, opened 2020-02-12
Restructure the template files

rosle:
## What happened

✅ Restructure the template files

⚒ Remove `Guard` as we don't usually use it.
 
## Insight

Restructure the whole template

Move the files in `rails_docker` and `shared` to its place.

From this PR, there will be 2 types of file

1) `.<ext>.tt` file - This is for templating. Those file with `.tt` will be executed as an ERB template before copied to the destination. - https://rubydoc.info/github/wycats/thor/master/Thor%2FActions:template

2) `.<ext>.rb` file - This is a Thor action file (like our main template). We will use this for modifying/adding the configurations to a file.

Reference: https://www.notion.so/Rails-Template-Re-engineering-4-0-0-90154ef2c1054d589e6bef4c1770101e 

## Proof Of Work

Spec should pass 👍 💯


template.rb
```diff
@@ -5,84 +5,52 @@
...
-      config.assets.compile = false
-
-    EOT
-  end
-
   after_bundle do
     run 'spring stop'
 
-    # Devise configuration
     generate 'devise:install'
```
olivierobert: One note is that this should be a variant later on too as we don't always need Devise. Same as for Webpack below ✌️

app/template.rb
```diff
@@ -0,0 +1,27 @@
...
+end
+
+# Stylesheets
+remove_file 'app/assets/stylesheets/application.css'
+directory 'app/assets/stylesheets'
+
+# Controllers
+directory 'app/controllers/concerns'
+inject_into_class 'app/controllers/application_controller.rb', 'ApplicationController' do
+  "  include Localization\n"
```
olivierobert: This extra space seems akward. How about to use the tab `\t`? or using two-character specs?

template.rb
olivierobert: Acutally for Devise, how about doing this in the main template file:

```
if yes?("Add devise?")
  gem 'devise'

  after_bundle do
    rails_command 'generate devise:install'
    rails_command "generate devise #{model}"
  end
end
```

So in the end, there would be pure two main variants "Web" and "API" and options like Devise, Boostrap, Figaro vs. DotEnv.

Webpack would still come by default with the Web variant.

What do you think?

rosle: I agree. Later we can add a different folders for that kind of options. 👍 

So for now I am going to leave the variants folder to have only `web` and `api`.

app/template.rb
rosle: Updated the test. I think later I would like to follow the [HEREDOC active_support layout](https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Layout/IndentHeredoc
) in the rubocop guide for the template text. 

Currently we're using `<<-EOT` for keeping the indent and `<<~EOT` for stripping the indent. But maybe we should use `<<-TEXT` and `<<-TEXT.strip_heredoc`  instead. So we always use dash. 🤔 

If you agree I can create the ticket in `4.0.0` and create a PR for it.

rosle:
> Not really related to this PR, but do we still need the following since the Semaphore CI 2.0 project has been deleted?
> 
> Screen Shot 2563-02-17 at 16 10 18

About the `smoke-test`, I'm not sure if it is related to Semaphore 2.0 🤔 I think it's just a test, I also try running that but it is not working as when starting the test image it will run the test not the web server. And currently seems like we use only make create and make run in our test.

Maybe we can remove both semaphore 2 and the smoke-test for now? Later when working on the test, we can revisit it 🤔 💭

app/template.rb
olivierobert: I agree. SO let's fix that in another PR.

PR #156, opened 2020-02-17
Create web variant

rosle:
## What happened

✅ Move web-specific gem + settings to `variants/web`

✅ Remove unused test (This will be revisited later when working on the test)

 - Remove Semaphore 2 setup
 - Remove smoke-test

✅ Add `eslint` and `scsslint` to codebase spec (on web variant)

✅  Add 2 tests on the semaphore, one for `web` and one for `api`

## Insight

To use api/web variant, create the template by:

Use only API
```
rails new <app_name> -m <template> --api
``` 
If the api option is not specified, it will apply web variant onto the app.

-------

1. Create the web variant, which is currently containing the setup for javascript, stylesheets and webpack setup.

2. Create the docker add-ons which contains only docker-related setup. On this add-on, we're using the ERB to templating the file as with `insert_into_file`, it is complicated to find the proper place to put the configurations in.

## Proof Of Work

Currently we added tests for each variant

Screen Shot 2563-02-19 at 11 57 08

All specs should pass 💚


.template/addons/docker/Dockerfile.tt
```diff
@@ -1,14 +1,18 @@
 FROM ruby:<%= RUBY_VERSION %>-slim
 
 ARG RUBY_ENV=development
+<%- if WEB_VARIANT -%>
 ARG NODE_ENV=development
+<%- end -%>
```
olivierobert: Why not using `insert_into_file Dockerfile` similarly as done for the Gemfile?

.template/addons/docker/docker-compose.test.yml.tt
```diff
@@ -2,27 +2,29 @@ version: '3.2'
...
-    image: redis:5.0.7
+    image: redis:<%= REDIS_VERSION %>
     container_name: <%= APP_NAME %>_redis
     ports:
       - "6379"
 
   test:
     build:
-      context: .
+      context: ../../..
```
olivierobert: This looks strange as it should build from the root directory 🤔

rosle: I think the rubymine automatically updated it after I move the file around. It should be `.` like the old one 👍

.template/addons/docker/Dockerfile.tt
rosle: For this one, I'm trying to make an addon for the docker as it is different between web and api. I am thinking about many ways that we could do for this one. 

1. Duplicate the files onto web and api, the con is that we need to maintain many files.
2. Use `insert_into_file`, but this one might be a lot to find where to insert, it is very scattered on the docker file. Going this way also make it hard to move docker out to addon. Apply it on the base would be the easiest as we can place the modifying thing on the variant folder.
3. Move docker to addon and use condition to modifying the file. (The current one) - This still feels weird as we usually do insert on the variant.

I think I can try to work on the insert thing to see how it looks like. 🤔 And we can discuss how should we do it.

addons/docker/Dockerfile.tt
```diff
@@ -1,16 +1,19 @@
...
-ARG BUILD_ENV=development
 ARG ASSET_HOST=http://localhost
+<%- end -%>
 
 # Define all the envs here
-ENV RACK_ENV=$RUBY_ENV \
-    RAILS_ENV=$RUBY_ENV \
+ENV BUILD_ENV=$BUILD_ENV \
+    RACK_ENV=$RUBY_ENV \
+    RAILS_ENV=$RUBY_ENV<%- if WEB_VARIANT -%> \
```
olivierobert: The `if` placement is a bit awkward but I think it's fine as it prevents to create another `RUN` line (thus another Docker image layer).

variants/web/config/environments/test.rb
```diff
@@ -0,0 +1,17 @@
...
+  EOT
+end
+
+insert_into_file 'config/environments/test.rb', before: %r{^end} do
+  <<-EOT
+
+  # Disable all animation during tests
+  config.middleware.use Rack::NoAnimations
+
+  # Do not fallback to assets pipeline if a precompiled asset is missed.
```
olivierobert: Typo: is miss**ing**.

variants/web/package.json.rb
```diff
@@ -0,0 +1,22 @@
...
+    "scripts": {
+      "lint": "eslint . --color",
+      "lint:fix": "eslint . --color --fix"
+    },
+  EOT
+end
+
+insert_into_file 'package.json', after: %r{"@rails/ujs":.+\n} do
+  <<~EOT
+    "i18n-js": "^3.0.11",
```
olivierobert: Use a fixed version: "3.0.11" (no ^)

olivierobert: Also what about updating to [3.6.0](https://github.com/fnando/i18n-js/releases/tag/v3.6.0)?

variants/web/template.rb
```diff
@@ -0,0 +1,35 @@
...
+  apply 'app/template.rb'
+  apply 'config/template.rb'
+  apply 'package.json.rb'
+
+  remove_turbolinks
+
+  after_bundle do
+    use_source_paths [__dir__]
+
+    # Fix the default rails template that does not put trailing commas
```
olivierobert: Which issues does that solve? Can't the template files be fixed instead?

rosle: For this one, the problem is with rails by default they're not using the trailing comma. But if we not doing this, it won't pass our eslint.

You can check on their template here -> https://github.com/rails/rails/blob/v6.0.2.1/railties/lib/rails/generators/rails/app/templates/app/javascript/packs/application.js.tt

variants/web/config/environments/test.rb
rosle: Fixed 👍 - https://github.com/nimblehq/rails-templates/pull/156/commits/5c4211e206043e00d468c3377e9a9aa401a4738b

variants/web/package.json.rb
rosle: Update and lock the version on the Gemfile as well 👍 - https://github.com/nimblehq/rails-templates/pull/156/commits/f466ec1fee7faf997d161e515eb0774b10df7c36

variants/web/template.rb
olivierobert: Noted

PR #157, opened 2020-02-19
Create api variant

rosle:
## What happened

✅ Create the api variant
 
## Insight

- Setup localization from the `Accept-Language` header

## Proof Of Work

Spec should pass 🌲


olivierobert:
How about adding a few more gems that we use like `fast_jsonapi` and `json_matchers` ?

PR #159, opened 2020-02-20
Add template tests

rosle:
## What happened

✅ Test if files are created based on the template

✅ Test if the contents are inserted correctly

## Insight

 Using [Serverspec](https://serverspec.org/) to spec the project created from the template. Checking the files inside the docker image if it is created and inserted correctly.

There are currently 3 types of specs:

1) docker image - check if the server is listening

2) files creation - for testing the main `template.rb`, mostly for copying files and directory, we need to check if the files are created correctly.

3) files insertion - for testing the template `.rb` files, mostly check for the content insertion if it is inserted.

------

##### Running tests

- Modify the `Makefile` to add the step for running the tests

- Add the command on Semaphore

```
make create
make build;
make test VARIANT=web
make run
```

## Proof Of Work

Spec should pass 📗


.template/addons/docker/Dockerfile.tt
```diff
@@ -31,7 +31,7 @@ ENV LANG="en_US.UTF-8" \
     LANGUAGE="en_US:en"
 
 RUN apt-get update -qq && \
-    apt-get install -y --no-install-recommends apt-transport-https curl gnupg
+    apt-get install -y --no-install-recommends apt-transport-https curl gnupg net-tools
```
rosle: Need to install this package to use `netstat` which is used by the serverspec to check if the port is listening - https://github.com/mizzy/specinfra/blob/8a19cf4840d2240e738c7f300e29fbe373f5fda7/spec/command/debian/port_spec.rb

Issue #148, opened 2020-02-10
Use webpacker pre-compilation in TEST ENV


Tolsee:
# Problem
- Webpacker compiles on demand rather than upfront by default
- System tests are taking longer to complete, because of on-demand asset compilation

# Solution
- Use webpacker pre-compilation

olivierobert:
@rosle let's include that in the major refactor for 4.0.0 🤘

PR #160, opened 2020-02-20
Webpack on test

Tolsee:
Closes https://github.com/nimblehq/rails-templates/issues/148 

## What happened

- Add webpack pre-compilation on TEST environment
- Add spec support to run the webpack server
 
## Insight

`N/A`

## Proof Of Work

- Configuration for webpack is added

Screen Shot 2563-03-13 at 17 59 22
 
- spec support for webpack is created

Screen Shot 2563-03-13 at 18 00 43

- Tests should pass


Tolsee:
@rosle @olivierobert 

Another important thing we need for this to work is setting `compile: false` in webpacker config
Reference: https://github.com/rails/webpacker#custom-rails-environments

How do you think we should do that?

rosle:
@Tolsee Normally if the file created by rails, we do modify the file with `insert_into_file`. 

We can create `config/webpacker.yml.rb` and then apply on the config/template.rb. 

main config template: https://github.com/nimblehq/rails-templates/blob/d29216dcef16f36cd7b6246d1b95af80a0a48f4c/config/template.rb

`insert_into_file` example: https://github.com/nimblehq/rails-templates/blob/d29216dcef16f36cd7b6246d1b95af80a0a48f4c/config/application.rb

PR #162, opened 2020-02-25
Add `how to contribute` section to README file

rosle:
## What happened

✅ Update the `README`

## Proof Of Work

Please check -> https://github.com/nimblehq/rails-templates/blob/chore/readme/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ For any changes made, you **must** a spec for it.
 
 ## License
 
-This project is Copyright (c) 2014-2019 Nimble. It is free software,
+This project is Copyright (c) 2014-2020 Nimble. It is free software,
 and may be redistributed under the terms specified in the [LICENSE] file.
 
 [LICENSE]: /LICENSE
--- a/README.md
+++ b/README.md
@@ -19,10 +19,71 @@ Install ruby and set your local ruby version to `2.6.5`
 The main template is `rails_docker`. In order to use it, initialize a new app with the following parameters:
 
 ```
-rails new <app_name> -m https://raw.githubusercontent.com/nimblehq/rails-templates/master/template.rb -T
+rails new <app_name> -m https://raw.githubusercontent.com/nimblehq/rails-templates/master/template.rb
 ```
 
+Supported template options:
+- `--api` - create an api-only application
+
 Read more about Rails Application Template in the [official Rails Guides](https://guides.rubyonrails.org/rails_application_templates.html).
+
+## How to contribute
+
+#### Template structure
+
+```
+.
+├── .template
+│   ├── addons
+│   │   └── docker
+│   │       ├── ...
+│   │       └── template.rb
+│   └── variants
+│       ├── api
+│       │   ├── ...
+│       │   └── template.rb
+│       └── web
+│           ├── ...
+│           └── template.rb
+├── app
+├── bin
+├── config
+├── spec
+├── ...
+├── README.md
+├── README.md.tt
+└── template.rb
+```
+
+We keep the Rails-app-like structure. On the root, there are base project file templates. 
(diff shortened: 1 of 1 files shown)


.template/addons/docker/Dockerfile.tt
```diff
@@ -31,7 +31,7 @@ ENV LANG="en_US.UTF-8" \
     LANGUAGE="en_US:en"
 
 RUN apt-get update -qq && \
-    apt-get install -y --no-install-recommends apt-transport-https curl gnupg
+    apt-get install -y --no-install-recommends apt-transport-https curl gnupg net-tools
```
olivierobert: What is `net-tools` for?

README.md
```diff
@@ -19,11 +19,72 @@ Install ruby and set your local ruby version to `2.6.5`
...
+├── README.md.tt
+└── template.rb
+```
+
+We keep the Rails-app-like structure. On the root, there are base project file templates. 
+Other files including the template options are in `.template` folder.
+
+There are 2 kinds of the template options:
+
+1. **Variants** - For the app main options, which is `web` and `api`.
```
olivierobert: English: which are

README.md
```diff
@@ -19,11 +19,72 @@ Install ruby and set your local ruby version to `2.6.5`
...
+    
+2. **Addons** - For other extra options that we can add to the project like `docker` or `bootstrap`. 
+The questions will be prompted before generating the project.
```
olivierobert: This needs to be in one sentence:

```
For other extra options that we can add to the project like `docker` or `bootstrap`, use the prompt `ask` to generate a question before generating the project.
```

Also technically this is no implemented yet. But can we put an example of how it would look like and also where it should find in the root template file?

README.md
```diff
@@ -19,11 +19,72 @@ Install ruby and set your local ruby version to `2.6.5`
...
+2. **`.rb` files**
+
+    This is used for modifying the files that Rails has generated.
+    The file name should be the same as on the generated app. 
+    If it is not a ruby file, append the `.rb` as an extension e.g. `Gemfile.rb`
+    
+#### Template specs
+
+We are using [Serverspec](https://serverspec.org/) to test the template.
+For any changes made, please add a spec for it. We also keep the same structure as the template.
```
olivierobert: How about being more directive:

```
For any changes made, you **must** add specs for it.
```

The second part of the sentence is not really needed as it's assumed from the first section `Template structure`

README.md
```diff
@@ -19,11 +19,72 @@ Install ruby and set your local ruby version to `2.6.5`
...
 ```
 
+Supported template options:
+- `--api` - create an api-only application
+
 Read more about Rails Application Template in the [official Rails Guides](https://guides.rubyonrails.org/rails_application_templates.html).
 
+## How to contribute
+
+#### Template structure
```
olivierobert: Why making an `h4` instead of an `h4` as the previous heading is an `h2`?

.template/addons/docker/Dockerfile.tt
rosle: We need to install this package to use netstat. The serverspec use this one to check if the port is listening.

README.md
```diff
@@ -19,11 +19,72 @@ Install ruby and set your local ruby version to `2.6.5`
...
+    If it is not a ruby file, append the `.rb` as an extension e.g. `Gemfile.rb`
+    
+### Template specs
+
+We are using [Serverspec](https://serverspec.org/) to test the template.
+For any changes made, you **must** a spec for it.
+
 ## License
 
 This project is Copyright (c) 2014-2019 Nimble. It is free software,
```
olivierobert: @rosle let's remove or update the date to match 2020

PR #163, opened 2020-02-27
Integrate with Semaphore 2

rosle:
## What happened

✅ Integrate with semaphore 2
 
## Insight

Setup the pipeline on Semaphore 2.0 to test the template.

## Proof Of Work

Screen Shot 2563-02-28 at 12 08 09

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -15,7 +15,6 @@ global_job_config:
       - checkout
       - sem-version ruby 2.6.5
       - sem-version node 12.16.1
-      - gem install rails -v 6.0.1 --no-document
       - export BRANCH_TAG=${SEMAPHORE_GIT_BRANCH/\//-}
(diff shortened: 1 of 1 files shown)


rosle:
@olivierobert This is working now. There are things that I am not sure and might need more input

About the pipeline, we could go with the simpler pipeline like -> https://nimble.semaphoreci.com/workflows/b061737d-00ae-407e-9dfe-228fa861f31c. With the simpler one, the yml file is much simpler. 

But with the current one (as in the screenshot), we have more steps and when it fails , it is easier to see which step is failed. But this also comes with more complicated configurations. You can check the config if it's too complicated or not 🤔 💭 

-----

For how we testing the addons. I'm still not sure about it. It's something we have to look further on. I feel like maybe after the web variant test is finish, we can just apply the addon template and run the addon template. 

With semaphore 2, we can do in parallel, something like

```
web variant test +-> apply addon 1 --> test addon 1
                 | 
                 +-> apply addon 2 --> test addon 2
```

But not quite sure yet. Just an idea 😄💡

PR #177, opened 2020-03-10
Semaphore CI 2.0 Addon

rosle:
## What happened

✅ Add semaphore addons
 
## Insight

The project will have a Semaphore CI 2.0 integrated as default. 

### Docker build problem for production

There is a known issue for building the production image. Because we precompile the assets in the build. It initializes the app. So all environment variables need to exist in the build stage because we always use `ENV.fetch` to fast-fail when there is no variables set.

So we're trying to find an easier way to do solve that. But for now, we can put all in the build arguments. This solves the problem but it could effect:

1. **Image Layers**

   For the layers, it will effect if we do the Heroku container deployment. As the Heroku sets the limit as 40 layers. So it will be failed to deploy when the image has more than 40 layers - https://devcenter.heroku.com/articles/container-registry-and-runtime#known-issues-and-limitations

    Also for the layers, we need to keep watching if the app has engines. For an engine, we need to add 2 more instruction to copy the gemfile for bundling.

    ```
    COPY engines/app_auth/lib/ ./engines/app_auth/lib/
    COPY engines/app_auth/Gemfile engines/app_auth/app_auth.gemspec ./engines/app_auth/
    ```

    For this one, I didn't find any solution for now. If we remove that and just do copy all, it could slow down the build as it needs to bundle every time the code changes. So we need to keep that in mind and live with it. Until we have another deployment option.

2. **Image Size**

    For the size, I checked but the size for ARG layer is 0, so I did not see any size changes.

## Proof Of Work

To test, I created a new app from the template and use the manual promotion to deploy to Heroku.

The app should be successfully deployed and running.

Screen Shot 2563-03-11 at 16 37 42

The final workflow

Screen Shot 2563-03-11 at 16 45 07


.template/addons/semaphore/.semaphore/semaphore.yml.tt
```diff
@@ -0,0 +1,66 @@
...
+          commands:
+            - artifact push job tmp/screenshots
+
+promotions:
+  - name: Staging
+    pipeline_file: deploy-staging.yml
+    auto_promote:
+      when: "branch = 'development' AND result = 'passed'"
+  - name: Production
+    pipeline_file: deploy-production.yml
```
olivierobert: Since it's called promotions, how about renaming these files as `promotion-production.yml` and `promotion-staging.yml`?

.template/addons/docker/Dockerfile.tt
```diff
@@ -120,6 +120,12 @@ RUN yarn install --network-timeout 100000
 COPY . ./
 
 <%- if WEB_VARIANT -%>
+# Set environment to development in order to precompile the assets
+# Because it initializes the app, so all variables need to be exist
```
olivierobert: English: so all variables need to ~~be~~ exist

.template/addons/docker/Dockerfile.tt
```diff
@@ -120,6 +120,12 @@ RUN yarn install --network-timeout 100000
 COPY . ./
 
 <%- if WEB_VARIANT -%>
```
olivierobert: Why are these ARG defines so low in the file and not with the other ARG on line 5-8?

PR #178, opened 2020-03-12
Enable single template application

rosle:
## What happened

- Allow to apply single addon onto an existing application
 
## Insight

The path needs to be fixed to make it work remotely.

To apply a single template, we need to call it through the main template as the repo needs to be cloned before it is templating.

```
rails app:template LOCATION=<template_url> ADDON=<addon name>
```

For an api application, there are no option provided like rails new command. So instead, you can pass `API` variable.

```
rails app:template LOCATION=<template_url> API=true
```

## Proof Of Work

You can try create/update the application with the template on this branch

```
rails new <app_name> -m https://raw.githubusercontent.com/nimblehq/rails-templates/feature/apply-single-template/template.rb

rails app:template LOCATION=https://raw.githubusercontent.com/nimblehq/rails-templates/feature/apply-single-template/template.rb

rails app:template LOCATION=https://raw.githubusercontent.com/nimblehq/rails-templates/feature/apply-single-template/template.rb API=true

rails app:template LOCATION=https://raw.githubusercontent.com/nimblehq/rails-templates/feature/apply-single-template/template.rb ADDON=docker
```

### DIFF
--- a/.template/addons/docker/template.rb
+++ b/.template/addons/docker/template.rb
@@ -1,4 +1,4 @@
-use_source_paths [__dir__]
+use_source_path __dir__
 
 template 'Dockerfile.tt'
 template 'docker-compose.dev.yml.tt'
--- a/.template/addons/semaphore/template.rb
+++ b/.template/addons/semaphore/template.rb
@@ -1,4 +1,4 @@
-use_source_paths [__dir__]
+use_source_path __dir__
 
 directory '.semaphore'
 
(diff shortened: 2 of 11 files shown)


rosle:
For applying the template on the existing project, there are 2 scenarios

1. Apply the whole template - For this we can just do `rails app:template LOCATION=<template_url>`, This will apply the base template + the addons. During this process, it will prompt the yes/no questions for addons like the normal create flow.

2. Apply only addon - For the project that might have some problem on upgrading, they can choose to apply only the addons they want or in priority. For this, they can pass the `ADDON=<addon_name>`, which it will apply only the addon template they selected. So in this case, it won't prompt the user for the addon.

olivierobert:
@rosle rebase required and let's run the updated tests on Semaphore 2

PR #182, opened 2020-03-17
Add precompiling assets script to fix docker build

rosle:
## What happened

✅ Add precompile assets script
 
## Insight

In order to precompile the assets inside the docker image, we need to have all variables set to make the app successfully initializes. This is because we always use ENV.fetch to get the environment variables to fast-failing when the env is not set properly. 

To fix, we add the docker precompile assets script. Which will take all variables needed and set the dummy values. So all variables is available for the app to initialize.

## Proof Of Work

- Spec should passed 🍏
- We can run the docker-compose build successfully

### DIFF
--- a/bin/docker-assets-precompile
+++ b/bin/docker-assets-precompile
@@ -1,0 +1,22 @@
+#!/usr/bin/env ruby
+
+# For building production docker image
+#
+# It sets the envs inside the docker image for precompiling the assets
+# Because to precompile the assets, Rails initializes the app.
+# And it requires the envs as we always use `ENV.fetch` to setup the variables
+#
+# Related issue: https://github.com/rails/rails/issues/32947
+
+require 'yaml'
+
+rails_env = ENV.fetch('RAILS_ENV', 'production')
+env_config = YAML.load_file('config/application.yml')[rails_env]
+
+if rails_env == 'production'
+  env_config.keys.each { |name| ENV[name] = 'dummy_value' }
+else
+  env_config.each { |name, value| ENV[name] = value }
+end
+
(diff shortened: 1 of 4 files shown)


config/application.yml.tt
```diff
@@ -13,3 +13,11 @@ development:
 test:
   <<: *default
   TEST_RETRY: "0"
+
+# Set environment variables required in the initializers in order to precompile the assets.
+# Because it initializes the app, so all variables need to exist in the build stage.
+production:
```
anduonghien, line 19: To avoid the confusion between this and the real Production, how about we rename this one to `build` 🤔 That means some env during the `build` step 🤔 

or `compile` or `pre_production` 🌩️

bin/docker-assets-precompile
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env ruby
```
ankitkalia1195, line 1: @rosle Since this is a ruby script, should the file name end in `.rb` ?

andyduong1920: I think we can go without the `.rb` @ankitkalia1195 as it won't make the confusion with the Ruby class

Also Rails don't use `.rb` in their `ruby script` as well, for example the `bin/setup` file

image

rosle:
LGTM 👍

bin/docker-assets-precompile
ankitkalia1195: @andyduong1920 Sure, got it

ankitkalia1195 (approved):

hoangmirs (approved):
LGTM

PR #183, opened 2020-03-17
Add env fetcher to the template to fix docker build

rosle:
## What happened

✅ Init the Env helpers for the application
 
## Insight

In order to precompile the assets inside the docker image, we need to have all variables set to make the app successfully initializes. This is because we always use `ENV.fetch` to get the environment variables to fast-failing when the env is not set properly.

To fix, we add the `env fetcher`, which will return a dummy value when we build the image. We already have something like this in one of the project where we want to fast-failing when fetching the credentials.

To retrieve the env variables:

```
App::Env.fetch('ENV_NAME')

# Cast to specific type
App::Env.fetch('ENV_NAME', type: :integer)

# With default value
App::Env.fetch('ENV_NAME', type: :integer, default: 2)
```

To retrieve the credentials

```
# To fetch
# development:
#   aws:
#     access_key_id: <my_access_key>

App::Env.fetch('aws.access_key_id', source: :credentials)
```

## Proof Of Work

- Spec should passed 🍏 
- We can run the `docker-compose build` successfully


.template/addons/docker/Dockerfile.tt
```diff
@@ -108,8 +103,8 @@ RUN rm -rf tmp/docker
 
 <%- if WEB_VARIANT -%>
 # Compile assets
-RUN bundle exec rails i18n:js:export && \
-    bundle exec rails assets:precompile
+RUN SECRET_KEY_BASE=1 DOCKER_BUILD=true bundle exec rails i18n:js:export && \
+    SECRET_KEY_BASE=1 DOCKER_BUILD=true bundle exec rails assets:precompile
```
olivierobert: How about moving that to a Shell script instead? The advantage is that we could document what DOCKER_BUILD is for and why SECRET_KEY_BASE is needed

```suggestion
RUN ./bin/assets-precompile
```

config/application.rb
```diff
@@ -13,5 +13,9 @@
   
     # Compress the responses to reduce the size of html/json controller responses.
     config.middleware.use Rack::Deflater
+
+    # Configure the autoload and eager load paths
+    config.autoload_paths << Rails.root.join('lib')
+    config.eager_load_paths << Rails.root.join('lib')
```
olivierobert: Eager loading the whole `lib` directory is usually bad practice. Instead it would be better to load `lib/<APP_DIRECTORY_NAME>` e.g. `lib/mia-web`

template.rb
```diff
@@ -1,9 +1,69 @@
...
+use_source_path template_root
+
 # Variables
-APP_NAME = app_name
+APP_NAME = @template_application.app_name
 # Transform the app name from slug to human-readable name e.g. nimble-web -> Nimble
-APP_NAME_HUMANIZED = app_name.split(/[-_]/).map(&:capitalize).join(' ').gsub(/ Web$/, '')
+APP_NAME_HUMANIZED =  @template_application.humanized_name
+APP_DIRECTORY_NAME = @template_application.directory_name
+APP_NAMESPACE = @template_application.namespace
```
olivierobert: Let's add more documentation for `APP_DIRECTORY_NAME` and `APP_NAMESPACE` so that we can understand similarly what was done for `APP_NAME_HUMANIZED`

template.rb
```diff
@@ -1,9 +1,69 @@
...
+require "#{template_root}/.template/lib/template/errors"
+
+@template_application = Template::Application.new(app_name)
+@template_errors = Template::Errors.new
+
+# Init the source path
+@source_paths ||= []
+use_source_path template_root
+
 # Variables
```
olivierobert: The variables are placed too far down now since lots of methods have been pasted above 🤔 It would be better to keep those at the top of the file.

template.rb
```diff
@@ -1,9 +1,69 @@
...
+  if (branch = __FILE__[%r{rails-templates/(.+)/template.rb}, 1])
+    Dir.chdir(tempdir) { git checkout: branch }
+  end
+
+  tempdir
+end
+
+def delete_test_folder
+  FileUtils.rm_rf('test')
+end
```
olivierobert: This could be externalized to a separate file -> `test/template.rb`. It would just remove it the test directory 🙃 Might sound silly but it would allow externalizing this concern just like for other directories, then it could allow later on to have a flag to use Minitest if needed.

template.rb
```diff
@@ -1,9 +1,69 @@
...
+    There are some errors when templating the application, Please fix them manually:
+
+    #{@template_errors}
+    #{'=' * 80}
+  EOT
+end
+
+# Setup the template root path
+# If the template file is the url, clone the repo to the tmp directory
+template_root = __FILE__ =~ %r{\Ahttps?://} ? remote_repository : __dir__
```
olivierobert: Which `tmp` directory? 🤔 Why is this needed?

.template/lib/template/application.rb
```diff
@@ -0,0 +1,26 @@
...
+      @app_name = app_name
+    end
+
+    # Transform the app name from slug to human-readable name e.g. nimble-web -> Nimble
+    def humanized_name
+      @humanized_name = app_name
+                          .split(/[-_]/)
+                          .map(&:capitalize)
+                          .join(' ')
+                          .gsub(/ Web$/, '')
```
olivierobert: `Web` could be stored as a constant:

```suggestion
                          .gsub(/ APP_NAME_SUFFIX$/, '')
```

template.rb
rosle: This is for when we use the template from the url like `rails new my-app -m https://raw.github...`. We need to clone the repo into local before using, so it will be clone to a tmp directory created by `Dir.mktmpdir`

Screen Shot 2563-03-18 at 14 34 06

template.rb
rosle: Yeah It feels quite weird to put here. The reason is it need to init from the class we required above. So we need to clare it before declaring the variables

APP_NAME_HUMANIZED =  `@template_application`.humanized_name

-----

I think I'm going to call `@template_application.humanized_name` directly instead of stored it in the variables to avoid this weird ordering 👍

template.rb
rosle: Added in the `Template::Application` class

andyduong1920:
Close as the `https://github.com/nimblehq/rails-templates/pull/182` implementation is simpler than this one, thanks @rosle

PR #186, opened 2020-05-29
[Chore] Rename makefile name and respect makefile tab indent style

namtx:
## What happened

- Rename `makefile` to `Makefile` as Makefile recommendation
- Respect Makefile tab indent style in .editorconfig

## Insight
 
- https://www.gnu.org/software/make/manual/html_node/Makefile-Names.html
- https://www.gnu.org/software/make/manual/html_node/Recipe-Syntax.html#Recipe-Syntax

## Proof Of Work

`N/A`


olivierobert:
Thanks for the contribution @namtx

PR #190, opened 2020-07-16
Prepare webpack assets only if there is at least one system test

hoangmirs:
## What happened
Prepare webpack assets only if there is at least one system test is executed

## Insight
Just change to use `when_first_matching_example_defined` config

## Proof Of Work
- Run normal tests (not type: :system) faster without webpack compilation
- The webpack compilation should be run once before running system tests

### DIFF
--- a/.template/variants/web/spec/support/webpack.rb
+++ b/.template/variants/web/spec/support/webpack.rb
@@ -1,5 +1,5 @@
 RSpec.configure do |config|
-  config.before(:suite) do
+  config.when_first_matching_example_defined(type: :system) do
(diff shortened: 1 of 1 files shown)


.template/variants/web/spec/support/webpack.rb
```diff
@@ -1,5 +1,5 @@
 RSpec.configure do |config|
-  config.before(:suite) do
+  config.when_first_matching_example_defined(type: :system) do
```
anduonghien, line 2: @hoangmirs Do you think we can add the `when_first_matching_example_defined` to the `/spec/support/assets.rb` file too 🤔

PR #207, opened 2020-08-18
Adding .nvmrc file

### LINKED ISSUE #167: Set the node version of the project using `.nvmrc`
Should we add `.nvmrc` file to the project to set the project node version for the web variant?

anduonghien:
## What happened
Resolve #167

 
## Insight
Using the latest LTS node version
 

## Proof Of Work
image


.template/spec/variants/web/template_spec.rb
```diff
@@ -9,6 +9,10 @@
     expect(file('.scss-lint.yml')).to exist
   end
 
+  it 'creates the .npmrc file' do
+    expect(file('.npmrc'))
```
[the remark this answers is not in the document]
anduonghien: Fixed that in https://github.com/nimblehq/rails-templates/pull/207/commits/41ab51f81b24fca722eb2b4d5ff5b6306b944998 @namtx

longnd (approved):

.template/addons/docker/Dockerfile.tt
```diff
@@ -23,7 +23,7 @@ ENV BUILD_ENV=$BUILD_ENV \
     <%- if WEB_VARIANT -%>
     ASSET_HOST=$ASSET_HOST \
     NODE_ENV=$NODE_ENV \
-    NODE_VERSION="12" \
+    NODE_VERSION=<%= NODE_SOURCE_VERSION %> \
```
namtx, line 26: How about change it to 
```suggestion
    NODE_SOURCE_VERSION=<%= NODE_SOURCE_VERSION %> \
```
?

namtx: it's kind of confused between NODE_VERSION and NODE_SOURCE_VERSION

anduonghien: Hi @namtx 

- NODE_VERSION is the nvm version (used for `.nvmrc`) 
- NODE_SOURCE_VERSION is the NodeSource Node.js Binary Distributions used to install Node in the Docker image.

Let me try to use the same value (12) for `nvm` if it works or not.

anduonghien: Setting the `.nvmrc` with `12` doesn't work as expected @namtx

image

namtx: I mean 
```
NODE_SOURCE_VERSION=<%= NODE_SOURCE_VERSION %>
```
it makes me confused because, in our template, we're using `NODE_VERSION` for assigning the node version but in Dockerfile, we're using it as `NODE_SOURCE_VERSION`

anduonghien: It's because

On development env, we are using nvm to control the Node version because we might have many different Node versions depend on each project, 

But in the Docker, we just need to use only 1 Node version so there is no point to setup `nvm` there, just install the Node directly.

Do you have other ideas to make it clear? 🤔

namtx: @anduonghien like I commented, how do we rename in NODE_VERSION to NODE_SOURCE_VERSION in Dockerfile?

namtx: https://github.com/nimblehq/rails-templates/pull/207/files#r476995350
the node version in the nvm file and Dockerfile is not related.

anduonghien: Oh got it @namtx I misunderstanding your comment before, changed it in https://github.com/nimblehq/rails-templates/pull/207/commits/5a57ba42d51002c0c8e84a45ee2b90f88c271d88

namtx (approved):

PR #212, opened 2020-08-18
Remove the spec/template.rb in the new app

### LINKED ISSUE #211: spec/template.rb appear in the generated app
### Why

The `spec/template.rb` still appear in the generated app

image

anduonghien:
## What happened
Resolve #211

 
## Insight
Just remove the file because the `directory 'spec', force: true` copied the whole rspec folder
 

## Proof Of Work
image

### DIFF
--- a/spec/template.rb
+++ b/spec/template.rb
@@ -1,6 +1,5 @@
 generate 'rspec:install'
 
-directory 'spec', force: true
-FileUtils.rm('spec/template.rb')
+directory 'spec', exclude_pattern: 'template.rb', force: true
 
 copy_file '.rspec', force: true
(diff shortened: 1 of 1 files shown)


spec/template.rb
```diff
@@ -1,5 +1,6 @@
 generate 'rspec:install'
 
 directory 'spec', force: true
```
hoangmirs, line 3: Can we pass the option `exclude_pattern: /pattern/` to directory action?
[That action supports](https://www.rubydoc.info/github/erikhuda/thor/master/Thor/Actions:directory) to exclude files with pattern

anduonghien: That works @hoangmirs , thanks for the suggestion, fixed it in https://github.com/nimblehq/rails-templates/pull/212/commits/1e4688bc43ae0a0803139306b6c06dd51ea5b64c

hoangmirs (approved):

bterone (approved):

PR #222, opened 2020-09-11
Fix Rubocop offense when first use template

gutakk:
## What happened
- When I start using this template, run test failed
- And when run `rubocop` command after `docker-prepare` things broken.
Screen Shot 2563-09-11 at 17 50 57

## Insight
- Update rubocop config to ignore tmp folder
- Combine loops in `vcr.rb` file according to rubocop error
 

## Proof Of Work
- When initialize a new app using this template, test will pass when you `run test`
- After `docker-prepare`, everything should work fine when you run `rubocop` command
Screen Shot 2563-09-11 at 18 17 10

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,6 +11,7 @@ AllCops:
     - 'Guardfile'
     - 'node_modules/**/*'
     - 'config/**/*'
+    - 'tmp/**/*'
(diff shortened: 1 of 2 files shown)


olivierobert (approved):
LOoks good to me

hoangmirs (approved):

PR #224, opened 2020-09-16
Integrate github actions

hoangmirs:
## What happened
- Integrate Github actions to test rails-template.
- Support different docker registries (docker.io, ghcr.io)
 
## Insight
- I've tried a lot with Github actions, so now I wanna integrate it to our template
- Create 2 workflows:
  + The first workflow is doing the same thing as Semaphore
  + The other is used for verifying production docker image. So we can avoid some mistakes when building images
- Support different registries by adding an environment variable named `DOCKER_REGISTRY`. Then modify the workflows using it
- Use [matrix feature](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategy) to check different variations of the template
- Move heroku stuff (Docker.web, Docker.worker) to another addon as we also use them in both semaphore and github action
- Create github addon. For now, it only includes `PULL_REQUEST_TEMPLATE.md` file. The main `.github` folder is using only for `rails-template` repo
- Use [this action](https://github.com/styfle/cancel-workflow-action) to cancel the previous runs, no need to run the redundant runs when we push new source code
- Use `actions/cache` to improve the speed (cache gems, npm packages,...)

## Proof Of Work
- Tests should be passed
- Tested by creating a new project from the template
- Work on [Github actions](https://github.com/nimblehq/rails-templates/actions)


.template/addons/semaphore/template.rb
```diff
@@ -1,6 +1,3 @@
 use_source_path __dir__
 
 directory '.semaphore'
-
-template 'Dockerfile.web.tt'
-template 'Dockerfile.worker.tt'
```
andyduong1920, line 6: Correct me if I'm wrong, I think we need to keep this as the `Dockerfile.web.tt` and the `Dockerfile.worker.tt` still there

hoangmirs: I moved it into Heroku addon, as we're not only using it on Semaphore but also on Github action

.github/workflows/variants_docker_build_test.yml
```diff
@@ -0,0 +1,90 @@
...
+            ${{ steps.yarn-cache-dir-path.outputs.yarn_cache_dir }}
+            **/yarn.lock
+            **/node_modules
+          key: ${{ runner.os }}-npm-packages-${{ hashFiles('**/package.json*') }}
+          restore-keys: |
+            ${{ runner.os }}-npm-packages-
+
+      - name: Create rails app
+        run: |
+          gem install rails -v ${{ env.RAILS_VERSION }} --no-document
```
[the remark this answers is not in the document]
hoangmirs: Make sense, I will update it

.github/workflows/variants_docker_build_test.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: Docker login
+        run: echo "$DOCKER_TOKEN" | docker login $DOCKER_REGISTRY --username=$DOCKER_USERNAME --password-stdin
+
+      - name: Docker pull
+        env:
+          BRANCH_TAG: ${{ env.BRANCH_TAG }}-${{ matrix.variant }}
+        run: |
+          cd $APP_NAME
+          docker-compose pull web || true
```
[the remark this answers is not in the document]
hoangmirs: Yes, we should. I changed it in https://github.com/nimblehq/rails-templates/pull/224/commits/9db661829eca41f87677142f1d7402e3e256f447

.template/addons/semaphore/template.rb
andyduong1920: so do we still need these files @hoangmirs 🤔

Pasted_Image_9_23_20__10_43

.semaphore/semaphore.yml
```diff
@@ -17,7 +17,7 @@ global_job_config:
       - sem-version node 12.16.1
       - export BRANCH_TAG=${SEMAPHORE_GIT_BRANCH/\//-}
       - export CACHE_NAME=$SEMAPHORE_GIT_BRANCH-$SEMAPHORE_PIPELINE_ID
-      - echo "$DOCKER_HUB_TOKEN" | docker login --username=$DOCKER_HUB_USERNAME --password-stdin
+      - echo "$DOCKER_TOKEN" | docker login $DOCKER_REGISTRY --username=$DOCKER_USERNAME --password-stdin
```
namtx, line 20: How about?
```suggestion
      - echo "$DOCKER_REGISTRY_TOKEN" | docker login $DOCKER_REGISTRY --username=DOCKER_REGISTRY_USERNAME --password-stdin
```

hoangmirs: Renamed in https://github.com/nimblehq/rails-templates/pull/224/commits/a69ce0df6b0519298835c98ac0650ec779d12cfd

.template/addons/semaphore/template.rb
hoangmirs: Deleted in https://github.com/nimblehq/rails-templates/pull/224/commits/a174f32b3cef4ec126c6cc05b4c518951b648910

.template/addons/github/.github/workflows/deploy.yml.tt
```diff
@@ -0,0 +1,64 @@
+name: Deploy
```
andyduong1920, line 1: @hoangmirs I think this file shoule belong to the next PR, is it?

andyduong1920 (approved):
LGTM @hoangmirs

namtx (approved):

olivierobert (commented):
I recommend keeping workflow naming with the pattern `action_component` e.g. test.yml, build.yml, deploy_staging.yml and deploy_production.yml. So here I think it would be more consistent to have `test_variants.yml` and `test_production_build.yml` 💭

.github/workflows/test.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: Gems cache
+        uses: actions/cache@v2
+        with:
+          path: /home/runner/.rubies/ruby-${{ env.RUBY_VERSION }}
+          key: ${{ runner.os }}-gems-${{ env.RUBY_VERSION }}-${{ hashFiles('**/Gemfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-gems-${{ env.RUBY_VERSION }}-
+
+      - name: Create app folder
+        if: ${{ matrix.variant == 'web' }}
```
andyduong1920, line 49: @hoangmirs Could you explain why we don't need to run this for API variant @hoangmirs

.github/workflows/test.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: Docker push
+        run: |
+          export BRANCH_TAG=${{ env.BRANCH_TAG }}-${{ matrix.variant }}
+          cd $APP_NAME
+          docker-compose push test
+
+      - name: Template test
+        run: |
+          export BRANCH_TAG=${{ env.BRANCH_TAG }}-${{ matrix.variant }}
+          make test_template VARIANT=${{ matrix.variant }}
```
andyduong1920, line 98: IMO, I prefer to move this test to a separate job, so we could separate the jobs to

- Template's unit test
- Variant test
- Production docker image build test

What do you think @hoangmirs

.github/workflows/test.yml
hoangmirs: I think I made this by mistake, Removed it

.github/workflows/test.yml
hoangmirs: With the current implementation, we still need to create a new app to run the template test, the same as variant test, so I will keep them together for now. We could separate them later when we refactor it

.github/workflows/test_production_build.yml
```diff
@@ -0,0 +1,78 @@
...
+
+on: push
+
+env:
+  APP_NAME: rails_templates
+  DOCKER_IMAGE: ${{ github.repository }}
+  DOCKER_REGISTRY_HOST: ${{ secrets.DOCKER_REGISTRY_HOST }}
+  RUBY_VERSION: 2.7.1
+  NODE_VERSION: 12
+  RAILS_VERSION: 6.0.3
```
andyduong1920, line 11: @hoangmirs Should we increase this one to 6.1.0 as the template has updated to Rails 6.1.0 in #263

.github/workflows/test_production_build.yml
```diff
@@ -0,0 +1,78 @@
+name: Test production build
+
+on: push
+
+env:
+  APP_NAME: rails_templates
+  DOCKER_IMAGE: ${{ github.repository }}
+  DOCKER_REGISTRY_HOST: ${{ secrets.DOCKER_REGISTRY_HOST }}
+  RUBY_VERSION: 2.7.1
+  NODE_VERSION: 12
```
andyduong1920, line 10: @hoangmirs Node 14 become current LTS version so it's ok to upgrade to NODE_VERSION 14, what do you think?

.github/workflows/test_production_build.yml
```diff
@@ -0,0 +1,78 @@
+name: Test production build
+
+on: push
+
+env:
+  APP_NAME: rails_templates
+  DOCKER_IMAGE: ${{ github.repository }}
+  DOCKER_REGISTRY_HOST: ${{ secrets.DOCKER_REGISTRY_HOST }}
```
andyduong1920, line 8: Do we need to define the `DOCKER_IMAGE` and `DOCKER_REGISTRY_HOST` here @hoangmirs 🤔 do we use it somewhere?

.template/addons/docker/docker-compose.test.yml.tt
```diff
@@ -19,14 +19,14 @@ services:
...
-        - ${DOCKER_IMAGE}:${BRANCH_TAG}
+        - ${DOCKER_REGISTRY_HOST}/${DOCKER_IMAGE}:${BRANCH_TAG}-test
       args:
         - BUILD_ENV=test
         - RUBY_ENV=test
         <%- if WEB_VARIANT -%>
         - NODE_ENV=test
         <%- end -%>
-    image: ${DOCKER_IMAGE}:${BRANCH_TAG}
+    image: ${DOCKER_REGISTRY_HOST}/${DOCKER_IMAGE}:${BRANCH_TAG}-test
```
longnd, line 29: I'm curious why we need to create a separated `test` image? in `docker-compose.yml.tt` we're naming the image as 
```
image: ${DOCKER_REGISTRY_HOST}/${DOCKER_IMAGE}:${BRANCH_TAG}
```
isn't it enough? what is the benefit of this test image? 🤔

.github/workflows/test_production_build.yml
hoangmirs: We don't need them here, they could be loaded from the env file, but I just want to make them consistent between workflows

.template/addons/docker/docker-compose.test.yml.tt
hoangmirs: The test image is often used for testing, the normal image is used for running the app.
Before, we usually use 1 name for both, but for me, we should create the separated images.
Both of them are pushed to the Docker registry, but if we use the same name, then 1 of them will be overridden by another.
So I think for each purpose, we should create different images.

andyduong1920:
@hoangmirs The CI is failed 😢

.github/workflows/test_production_build.yml
andyduong1920: 🤔  as we don't use it I think we should remove it to avoid confusion for the other contributor, also as you mentioned that `want to make them consistent` so how about the other env like

```
DOCKER_REGISTRY_USERNAME: ${{ github.repository_owner }}
DOCKER_REGISTRY_TOKEN: ${{ secrets.DOCKER_REGISTRY_TOKEN }}
```

As it's not in this workflow yet

.template/addons/docker/docker-compose.test.yml.tt
longnd: I think `docker-compose.test` is used for building test images (feature branches) while `docker-compose.yml` is used for production image (develop/master branch)
so even though they share the same naming pattern 
```yml
image: ${DOCKER_REGISTRY_HOST}/${DOCKER_IMAGE}:${BRANCH_TAG}
```
the images will never be overridden, IMO

.template/addons/bootstrap/application.js.rb
```diff
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
-insert_into_file 'app/javascript/packs/application.js', before: %r{import 'translations/translations'.+\n} do
+insert_into_file 'app/javascript/packs/application.js', after: %r{import "channels"\n} do
```
andyduong1920, line 3: @hoangmirs be careful as the previous implement we would like to put `vendor/bootstrap` under `core-js` stuff

image

andyduong1920: For that reason, I suggest removing this change and keeping the previous implementation @hoangmirs

.template/variants/web/template.rb
```diff
@@ -42,11 +42,13 @@ def remove_turbolinks
   end
 
   if File.exist?('app/javascript/packs/application.js')
-    gsub_file('app/javascript/packs/application.js', %r{^require\(\"turbolinks\"\).start\(\)\n}, '')
+    gsub_file('app/javascript/packs/application.js', %r{import Turbolinks from "turbolinks"\n}, '')
+    gsub_file('app/javascript/packs/application.js', %r{Turbolinks.start\(\)\n}, '')
```
andyduong1920, line 46: @hoangmirs Please rebase again, this was merged into the `develop` branch 😀

longnd (approved):

PR #230, opened 2020-09-24
Add rails-controller-testing to Gemfile template

gutakk:
## What happened
I tried to assert template in my Internal Certification. For example in this code
```ruby
describe "GET new" do 
  it "renders the new view" do 
    get :new

    expect(response).to render_template(:new)
  end
end
```
I got an error
Screen Shot 2563-09-24 at 16 15 11

So `rails-controller-testing` allow us to use `assigns` and `assert_template`.

## Insight
- Add `rails-controller-testing` to Gemfile template

Note : @hoangmirs concern is if there is another alternative or not because the last release of this lib was 4 years ago.
 

## Proof Of Work
- Test case above is passed

### DIFF
--- a/Gemfile.tt
+++ b/Gemfile.tt
@@ -61,6 +61,7 @@ group :test do
   gem 'simplecov', require: false # code coverage analysis tool for Ruby
   gem 'vcr' # Gem for recording test suite's HTTP interactions
   gem 'timecop' # Gem for time travel
+  gem 'rails-controller-testing' # Gem that allow to use assigns as well ass assert_template
 end
 
 group :production do


olivierobert:
@gutakk Indeed, the gem is quite old and to be honest this kind of test does not bring much value. For controller tests, usually, it's best to just test for HTTP status code (mainly `200`, `403`, `404`). Testing template rendering and assigns is kind of redundant with system tests. Also, especially testing assign is testing a detail of implementation instead of behaviour. 

💡 For view variables, make sure to use [local variables](https://compass.nimblehq.co/development/code-conventions/ruby-on-rails/#views) instead of variable instances.

andyduong1920 (approved):
LGTM @gutakk 👍

andyduong1920:
@olivierobert @hoangmirs @gutakk The gem isn't an old gem at all 😀  the latest release was on this Jun 23, haha

Pasted_Image_10_7_20__16_31

For me, it's ok to have this gem as it helps to test some simple stuff like `render_template(:new)`, it could help in testing unhappy path case (as we don't test the unhappy path on system test)

longnd (approved):

Issue #202, opened 2020-08-17
PhraseApp addon


andyduong1920:
### Why

We are using PhraseApp for localization so should we add PhraseApp to our template 

### Example

1/ Phrase app configuration

```
phraseapp:
  project_id: bdea2635d5497b473439bf269915c85e
  file_format: yml
  push:
    sources:
    - file: ./config/locales/<tag>.<locale_name>.yml
      params:
        skip_upload_tags: true
  pull:
    targets:
    - file: ./config/locales/<tag>.<locale_name>.yml
      params:
        include_empty_translations: true
        tags: tag_name
    ...
```

Issue #168, opened 2020-02-27
Test the template with Rubocop


rosle:
Add rubocop to test the template code.

PR #231, opened 2020-09-24
PhraseApp Addon

andyduong1920:
## What happened

Solved #202

At Nimble we used Phrase App in most projects, so better to have this Addon so we no need to spend time to set it up.
 
## Insight
The Phrase App addon is an optional add-on, they user can choose to have it or not on the generate new project.

The PhraseApp addon provides the `.phraseapp.yml` and the `spec` to check the PhraseApp pull configuration.

## Proof Of Work

1/ Question

Screen Shot 2020-09-24 at 16 46 23

2/ `.phraseapp.yml`

image

3/ Spec to check Phrase App pull configuration 

image


template.rb
```diff
@@ -59,7 +59,12 @@ def apply_template!(template_root)
   # Add-ons - [Default]
   apply '.template/addons/docker/template.rb'
   apply '.template/addons/semaphore/template.rb'
-
+  
+  # Add-ons - [Optional]
+  if yes?("Would you like to add PhraseApp configuration?")
```
rafayet-monon, line 64: Can we make it a one-liner and add a `(yes/no)` to make it more user friendly?

`apply '.template/addons/phrase_app/template.rb' if yes?("Would you like to add PhraseApp configuration?(yes/no)")`

Makefile
```diff
@@ -1,5 +1,6 @@
+# Y - in response to Would you like to add PhraseApp configuration?
 create:
-	rails new $(APP_NAME) -m ./template.rb -T ${OPTIONS}
+	printf "Y\n" | rails new $(APP_NAME) -m ./template.rb -T ${OPTIONS}
```
rafayet-monon, line 3: A reminder of `create_web` and `create_app` 😅

template.rb
rafayet-monon: Also, should we move it before `after_bundle` so we won't have to use after_bundle in [.template/addons/phrase_app/template.rb](https://github.com/nimblehq/rails-templates/pull/231/files#diff-18847e9c4ee2991b23f88d42738a1738). What do you think 🤔

.template/addons/phrase_app/template.rb
```diff
@@ -0,0 +1,6 @@
+after_bundle do
```
rafayet-monon, line 1: Do we need the `after_bundle` here as we are just copying and inserting into file?

.template/addons/phrase_app/spec/codebase/codebase_spec.rb
```diff
@@ -0,0 +1,21 @@
+insert_into_file 'spec/codebase/codebase_spec.rb', before: %r{end\Z} do
+  <<-EOT
```
rafayet-monon, line 2: We always use `EOT` as heredoc but rubocop always cries for a [meaningful heredoc naming](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/HeredocDelimiterNaming). Also adding the magic comment `frozen_string_literal: true` at the start of these types of files. Should follow these conventions?

Makefile
andyduong1920: Thanks for the reminder @rafayet-monon :D As we don't know which PR can merge first so once your PR is merged first I will rebase this one :D

.template/addons/phrase_app/template.rb
andyduong1920: yeah I tried without after_bundle and there is some error relative to the `spec` as I remember

template.rb
andyduong1920: I'm just wondering that if we make it in 1 line, the line could be really long.

rafayet-monon: If it's really long we should keep it as it is 👍

.template/addons/phrase_app/spec/codebase/codebase_spec.rb
andyduong1920: Hi @rafayet-monon Currently in the template we didn't have Rubocop set up yet, and many files are using `EOT` right now, also for the `frozen_string_literal: true` only 1 file having that 😃 so we will following the current template codebase @rafayet-monon 

And all will fix when we add the rubucop to the template itself https://github.com/nimblehq/rails-templates/issues/168

template.rb
andyduong1920: Refactor that in https://github.com/nimblehq/rails-templates/pull/231/commits/14e04aa961b618654cb3636bbad31be030676dd9 that will solve another issue which is the message template look like the same between each addon, right now each PRs has a different message structure

rafayet-monon (approved):
LGTM 🎉

template.rb
```diff
@@ -68,6 +71,10 @@ def apply_template!(template_root)
   apply '.template/hooks/before_complete/template.rb'
 end
 
+def install_addon_message(addon)
```
noor-ahmed, line 74: @andyduong1920 how does it sound? `installation_addon_message(addon)` ? 🦊

noor-ahmed (approved):
LGTM 💯

template.rb
```diff
@@ -100,6 +103,10 @@ def delete_test_folder
   FileUtils.rm_rf('test')
 end
 
+def install_addon_prompt(addon)
+  "Would you like to add #{addon} addon? [Yn]"
```
byhbt, line 107: Add the article `the` for the specific addon.

```suggestion
  "Would you like to add the #{addon} addon? [Yn]"
```

.template/spec/addons/base/phrase_app/template.rb
```diff
@@ -0,0 +1,9 @@
+describe 'PhraseApp addon - template' do
+  it 'creates .phraseapp.yml file' do
+    expect(file('.phraseapp.yml')).to exist
+  end
+  
+  it 'addings PhraseApp to codebase_spec' do
```
byhbt, line 6: `adds` seem more correct than `addings`

```suggestion
  it 'adds PhraseApp to codebase_spec' do
```

byhbt (approved):

PR #232, opened 2020-09-24
Nginx addon

andyduong1920:
## What happened
Solved #217 

Adding the Nginx addon, so we can save the time for adding the basic Nginx configuration

 
## Insight
Same as Phrase App, the Nginx addon is optional, the user can choose to generate it or not.

The Nginx addon changes:

- Dockerfile: Install Nginx, create the Nginx config for the application
- Start the Nginx along with the app

I used gsub_file  `config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?` and `config.public_file_server.enabled = true` cause 
- The `config.public_file_server.enabled = true` will happen in case the user applies this ADDON to the existing project. 
- The `config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?` apply to the project isn't gen by Nimble template.

```
gsub_file(
  'config/environments/production.rb',
  "config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?",
  'config.public_file_server.enabled = false',
)

gsub_file(
  'config/environments/production.rb',
  "config.public_file_server.enabled = true",
  'config.public_file_server.enabled = false',
)
```

## Proof Of Work
1/ Question

Screen Shot 2020-09-24 at 16 46 34

2/ Docker file

image

3/ Nginx configuration file

image


.template/addons/nginx/config/environments/production.rb
```diff
@@ -0,0 +1,11 @@
+gsub_file(
+  'config/environments/production.rb',
+  "config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?",
+  'config.public_file_server.enabled = false',
+)
+
+gsub_file(
```
gutakk, line 7: Do we really need this `gsub_file`? As if we decide to use nginx addon then `config.public_file_server.enabled = true` will not exists.

What do you think?

andyduong1920: it comes from the previous step @gutakk  https://github.com/nimblehq/rails-templates/pull/242

Here is the flow on generating the new rails project from template

```
# base config
https://github.com/nimblehq/rails-templates/pull/242

# addon
# nginx - optional...
```

gutakk: then should we only have this?

```ruby
gsub_file(
  'config/environments/production.rb',
  "config.public_file_server.enabled = true",
  'config.public_file_server.enabled = false',
)
```

IMO, if the the flow genere base config before nginx addon then  `config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?` will not exist as it is replaced by https://github.com/nimblehq/rails-templates/pull/242.

Please correct me if I'm wrong. I'm not 100% sure about this.

andyduong1920: The `config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?` will happend in case the user apply this ADDON to an existing project @gutakk 

We had [that option](https://github.com/nimblehq/rails-templates#use-the-template), the project isn't gen by our template so that might content the `config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
` which come from the Rails default one.

andyduong1920: I updated the PR description @gutakk, please have a look on it again.

gutakk: Thanks for explanation.

.template/addons/nginx/config/environments/production.rb
```diff
@@ -0,0 +1,11 @@
...
+  'config/environments/production.rb',
+  "config.public_file_server.enabled = true",
```
byhbt, line 9: Please update the quote to a single quote to make it consistent with the others.

andyduong1920: Fixed in https://github.com/nimblehq/rails-templates/pull/232/commits/40fe6a21861ff609f5f99ade75cd0958daaacc46 @byhbt

.template/addons/nginx/config/nginx/app.conf.template.tt
```diff
@@ -0,0 +1,113 @@
...
+    text/x-cross-domain-policy;
+
+  upstream app {
+    server localhost:3000;
+  }
+
+  server {
+    server_name _;
+
+    root /<%= APP_NAME %>/public;
```
byhbt, line 51: is the `<%= APP_NAME %>` good example for Nginx config?

Normally I use this one:

```suggestion
    root /var/www/html/<%= APP_NAME %>/public;
```

Or

```suggestion
    root /srv/apps/<%= APP_NAME %>/public;
```

.template/addons/nginx/config/nginx/app.conf.template.tt
```diff
@@ -0,0 +1,113 @@
...
+
+    proxy_set_header  X-Real-IP  $remote_addr;
+    proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
```
byhbt, line 54:
```suggestion
    proxy_set_header X-Real-IP  $remote_addr;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
```

Remove unnecessary 2 spaces.

andyduong1920: @byhbt could you explain more about `the alignment of the space is not consistent in this files` 🤔  at which line the alignment isn't consistent

.template/addons/nginx/config/nginx/app.conf.template.tt
andyduong1920: normally we deploy the Rails app to the root folder @byhbt it's different with PHP as the app locates in `var/www/html`.

byhbt: ok. LGTM

.template/addons/nginx/config/nginx/app.conf.template.tt
byhbt:
```
      proxy_buffer_size          12k;
      proxy_buffers              8 12k;
      proxy_busy_buffers_size    24k;
```
I think we can update it to: 😄 

```
      proxy_buffer_size 12k;
      proxy_buffers 8 12k;
      proxy_busy_buffers_size 24k;
```

byhbt (approved):

rafayet-monon (approved):
Looks Good 🎉

noor-ahmed (approved):

gutakk (approved):

.template/addons/nginx/config/nginx/app.conf.template.tt
```diff
@@ -0,0 +1,113 @@
+worker_processes auto;  ## Default: 1
+
```
ankitkalia1195, line 2: @andyduong1920 Should we have 2 versions of template, 1 for api and 1 for web? Because api will not need gzip and static file settings?

andyduong1920: Fixed in https://github.com/nimblehq/rails-templates/pull/232/commits/e770d2c22141fd6c4ffc2a5e450e4b69b854cff5 @ankitkalia1195

ankitkalia1195 (approved):

Issue #218, opened 2020-09-04
Heroku addon


andyduong1920:

ankitkalia1195:
@andyduong1920 Closing this issue as the description is not sufficient.

PR #238, opened 2020-10-02
Add Heroku Add-on

noor-ahmed:
## What happened
Solved https://github.com/nimblehq/rails-templates/issues/218
 
## Insight

- [x] Added add-on option to `template.rb`

If Dev wants to add Heroku, then
- [x] Added Heroku Configuration to `production.rb`
- [x]  Added variables to `.env`

## Proof Of Work
Screen Shot 2020-10-09 at 7 03 24 PM

### DIFF
--- a/.template/addons/heroku/config_template.rb
+++ b/.template/addons/heroku/config_template.rb
@@ -1,7 +1,5 @@
 insert_into_file 'config/environments/production.rb', before: /%r{end\nend\Z}/ do
   <<~CODE
-    config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
-
     if ENV["RAILS_LOG_TO_STDOUT"].present?
       logger           = ActiveSupport::Logger.new(STDOUT)
       logger.formatter = config.log_formatter
--- a/.template/addons/heroku/config_template.rb
+++ b/.template/addons/heroku/config_template.rb
@@ -1,0 +1,11 @@
+insert_into_file 'config/environments/production.rb', before: /%r{end\nend\Z}/ do
+  <<~CODE
+    config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
+
+    if ENV["RAILS_LOG_TO_STDOUT"].present?
+      logger           = ActiveSupport::Logger.new(STDOUT)
+      logger.formatter = config.log_formatter
+      config.logger = ActiveSupport::TaggedLogging.new(logger)
+    end
+  CODE
+end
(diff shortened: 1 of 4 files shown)


.template/addons/heroku/config_template.rb
```diff
@@ -0,0 +1,11 @@
+insert_into_file 'config/environments/production.rb', before: /%r{end\nend\Z}/ do
+  <<~CODE
+    config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
```
gutakk, line 3: I think we don't need this line. As we had [this](https://github.com/nimblehq/rails-templates/pull/242)

.template/addons/heroku/env_variables.rb
```diff
@@ -0,0 +1,2 @@
+echo 'RAILS_SERVE_STATIC_FILES=true' >> '.env' # default value
```
gutakk, line 1: Same as this [comment](https://github.com/nimblehq/rails-templates/pull/238#discussion_r503026561)

template.rb
```diff
@@ -60,6 +60,12 @@ def apply_template!(template_root)
   apply '.template/addons/docker/template.rb'
   apply '.template/addons/semaphore/template.rb'
 
+  if ENV['CI']
+    apply '.template/addons/heroku/template.rb'
+  elsif yes?("Would you like to add heroku?")
```
gutakk, line 65: How about add an action to guide user?

something like this

```ruby
Would you like to add Heroku? (yes/no)
```

.template/addons/heroku/config_template.rb
noor-ahmed: @gutakk Thanks for the good catch, resolved https://github.com/nimblehq/rails-templates/pull/238/commits/fb49d58544d911d81c69b8da4041687dfa85a460 🤝

.template/addons/heroku/env_variables.rb
noor-ahmed: resolved https://github.com/nimblehq/rails-templates/pull/238/commits/fb49d58544d911d81c69b8da4041687dfa85a460

template.rb
noor-ahmed: @gutakk yeah that I have also in my mind. We will make it an one line once reviewing done and ready for merging ✌️

ankitkalia1195:
@andyduong1920 Should I close this PR, as it is opened since long time and @noor-ahmed no longer works at the company. Also in this PR, I dont see anything related to heroku 🤔 . The issue you opened also does not contain any description https://github.com/nimblehq/rails-templates/issues/218.

andyduong1920:
> @andyduong1920 Should I close this PR, as it is opened since long time and noor no longer works at the company. Also in this PR, I dont see anything related to heroku 🤔 . The issue you opened also does not contain any description #218.

Yeah sure @ankitkalia1195 👍

Issue #165, opened 2020-02-27
Bootstrap addon


rosle:
Implement the addons for Bootstrap

The template should ask if the user wants to use Boostrap in the project

If yes, it should add the bootstrap configurations onto the project.

rosle:
Related with https://github.com/nimblehq/rails-templates/issues/123

Issue #123, opened 2019-10-14
Add Bootstrap CSS and Bootstrap Native JS


andyduong1920:
## Issue

We are using [Bootstrap CSS](https://getbootstrap.com/) and [Bootstrap Native JS](https://github.com/thednp/bootstrap.native) in most of the projects, should we add that to our template as well 🤔 

## Vote

[![](https://api.gh-polls.com/poll/01DQ4AHHHMS4NCH1N3FSF59F6R/Yes%2C%20let%20add%20the%20Bootstrap%20when%20we%20creating%20the%20new%20project)](https://api.gh-polls.com/poll/01DQ4AHHHMS4NCH1N3FSF59F6R/Yes%2C%20let%20add%20the%20Bootstrap%20when%20we%20creating%20the%20new%20project/vote)
[![](https://api.gh-polls.com/poll/01DQ4AHHHMS4NCH1N3FSF59F6R/No%20need%2C%20we%20can%20spend%20a%20few%20hours%20to%20add%20it%20manually%20later)](https://api.gh-polls.com/poll/01DQ4AHHHMS4NCH1N3FSF59F6R/No%20need%2C%20we%20can%20spend%20a%20few%20hours%20to%20add%20it%20manually%20later/vote)

PR #239, opened 2020-10-05
Add Bootstrap addon feature

rafayet-monon:
## What happened
Solved #165 & #123  

At Nimble we use Bootstrap for most of the project. In this PR we added Bootstrap as an add-on to minimize the project setup time.
 
## Insight

- [x] Add [Bootstrap](https://getbootstrap.com/)
- [x] Add [Bootstrap native](https://github.com/thednp/bootstrap.native)

- We added bootstrap native to `package.json`, then added a file to import bootstrap native and export it's components. The added the file to `application.js`. We used the latest `bootstrap.native - 3.0.13`
- We added bootstrap SCSS stylesheets to `vendor` and imported it as dependency in `application.scss`. We used latest `bootstrap - 4.5.2` for the stylesheets.
 

## Proof Of Work

- Ask while creating the project - 

image

- Bootstrap SCSS files used from the latest stable 

rsz_screenshot_2020-10-09_at_61454_pm

- `bootstrap.native` import and exporting modules

image


Makefile
```diff
@@ -11,14 +13,15 @@ run:
 	docker-compose -f docker-compose.test.yml run test
 
 base_spec = spec/base/**/*_spec.rb,spec/addons/docker/**/*_spec.rb
+bootstrap_spec = spec/addons/bootstrap/**/*_spec.rb
```
andyduong1920, line 16: How about we refactor here a bit @rafayet-monon 

```
base_addon_spec = spec/addons/**/*_spec.rb
web_addon_spec = spec/addons/variants/web/**/*_spec.rb
api_addon_spec = spec/addons/variants/api/**/*_spec.rb
base_spec = spec/base/**/*_spec.rb
web_spec = spec/variants/web/**/*_spec.rb
api_spec = spec/variants/api/**/*_spec.rb
...
	if [ $(VARIANT) = web ]; then \
	  bundle exec rspec --pattern="$(base_spec),$(web_spec)$(base_addon_spec),$(web_addon_spec),"; \
	elif [ $(VARIANT) = api ]; then \
	  bundle exec rspec --pattern="$(base_spec),$(api_spec)$(base_addon_spec),$(api_addon_spec),"; \
	fi;
```

Makefile
```diff
@@ -1,5 +1,7 @@
+# Y - in response to Would you like to add Bootstrap(yes/no)
+
 create:
-	rails new $(APP_NAME) -m ./template.rb -T ${OPTIONS}
+	printf "Y\n" | rails new $(APP_NAME) -m ./template.rb -T ${OPTIONS}
```
andyduong1920, line 4: @rafayet-monon as this response happened only on applying web variant, so how about we separate the `create` here to `create_web` and `create_api`

```
# Y - in response to Would you like to add Bootstrap(yes/no)
create_web:
  printf "Y\n" | rails new $(APP_NAME) -m ./template.rb -T --web ${OPTIONS}

create_api:
  rails new $(APP_NAME) -m ./template.rb -T --api ${OPTIONS}
```

Makefile
andyduong1920: so it required you move the `boostrap addon` from `.template/addons/boostrap` to `.template/addons/variants/web/boostrap`

andyduong1920: Or I can do the refactor on another PR, so you just need to  change the path to `.template/addons/variants/web/boostrap` @rafayet-monon

andyduong1920: The Makefile was refactored in https://github.com/nimblehq/rails-templates/pull/248, so please rebase this PR @rafayet-monon 🙏

Makefile
rafayet-monon: Added in [d534b0d](https://github.com/nimblehq/rails-templates/pull/239/commits/d534b0d6ffef858976ae1c54b9db07701fbaca71)

Makefile
rafayet-monon: Added in [d534b0d](https://github.com/nimblehq/rails-templates/pull/239/commits/d534b0d6ffef858976ae1c54b9db07701fbaca71)

andyduong1920 (approved):
LGTM @rafayet-monon 🙌

.template/addons/bootstrap/application.scss.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+insert_into_file 'app/assets/stylesheets/application.scss', after: /@import 'variables'.+\n/ do
```
gutakk, line 3: What do you think to import bootstrap in `Dependencies` section?
IMO bootstrap is dependencies instead of variables.

.template/addons/bootstrap/application.js.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+insert_into_file 'app/javascript/packs/application.js', after: %r{import 'core-js/stable'.+\n} do
+  <<~JAVASCRIPT
+    import 'vendor/bootstrap';
```
gutakk, line 5: Can we just `require('bootstrap.native')` instead of create our own boostrap native js file and import it?
What do you think?

.template/addons/bootstrap/template.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+use_source_path __dir__
+
+apply '.template/addons/bootstrap/application.scss.rb'
+apply '.template/addons/bootstrap/.scss_lint.yml.rb'
+apply '.template/addons/bootstrap/package.json.rb'
+apply '.template/addons/bootstrap/application.js.rb'
+
+directory('stylesheets/bootstrap', 'vendor/assets/stylesheets/bootstrap')
```
gutakk, line 10: Could you please explain more about why you decide to create bootstrap stylesheet ourselves instead of install bootstrap.
Thank you.

.template/addons/bootstrap/package.json.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+insert_into_file 'package.json', after: /"i18n-js":.+\n/ do
```
gutakk, line 3: How about add this after `"dependencies"`? If `i18n-js` get removed in any case I'm not sure if insertion will still work.

.template/addons/bootstrap/application.scss.rb
rafayet-monon: We could do that but the problem is I couldn't find a way where `insert_into_file` can detect the `Dependencies` comment and insert after that. I could add `// Dependencies` here but that would make it redundant having two `// Dependencies` in a file. It was bothering me too. What are your thoughts? 🤔

.template/addons/bootstrap/application.js.rb
rafayet-monon: The conventions that we follow in our projects is that we don't directly require third-party modules to the main file, in this case, `application.js`. We create a folder with respect to the name of the third-party plugin in `vendor` and import them in the main file. This gives us a certain level of independence if we want to override anything from the third-party and thus the `application.js` stays clean.

cc. @andyduong1920 is my explanation okay?

andyduong1920: First of all, we never override the third-party directly @rafayet-monon :D 

And the reason we put the bootstrap in vendor was it comes from a previous project, I don't set it up but me main use case is it's easy to look on that source code in order to customize something on my side.

Another thing, as the SCSS isn't handle by Webpakc in Rails, it's handle by Sprocket so I think at that time, people choose the Rails way, putting the scss to vendor folder or using the boostrap gem,

So it's a good time to test that again, @rafayet-monon could you spend time to try another solution

- Manage boostrap package with `package.json`
- Import the boostrap.scss from in the application.scss from the `node_modules/boostrap/` 

cc @gutakk

.template/addons/bootstrap/template.rb
rafayet-monon: Changed the implementations in [da9cb2f](https://github.com/nimblehq/rails-templates/pull/239/commits/da9cb2fda5a12c9cf2c6879281d8bc3044832bbb) as per the [comment](https://github.com/nimblehq/rails-templates/pull/239#discussion_r502198123).

Let me know your thoughts.

cc. @andyduong1920

.template/addons/bootstrap/application.scss.rb
gutakk: Would not be good if we have 2 `Dependencies` comment. I have no idea either about how to detect it.
@andyduong1920 @olivierobert Do you have any idea?

.template/addons/bootstrap/stylesheets/bootstrap/bootstrap.scss.tt
```diff
@@ -0,0 +1,42 @@
+// By default every component is imported
+// But DO NOT import the whole framework but instead
+// pick what the project requires
+// and comment out the rest.
+
```
gutakk, line 5: Can we just `@import 'bootstrap/scss/bootstrap'`? since in bootstrap.js file import all of this already.

Or not sure if we can `@import 'bootstrap/scss/bootstrap'` in `application.scss` directly?

.template/addons/bootstrap/template.rb
gutakk: I commented my thought [here](https://github.com/nimblehq/rails-templates/pull/239#discussion_r502236835)

.template/addons/bootstrap/stylesheets/bootstrap/bootstrap.scss.tt
rafayet-monon: Of course we could do that. But if you look at the comment in top of the file and the rule we have regarding using frameworks in [Compass](https://compass.nimblehq.co/development/code-conventions/css/#frameworks) -

> But frameworks also have downsides such as file size and unused code among others. So to protect the application from these downsides, our usage of frameworks comes with restrictions and conventions.
> 
> Do NOT import the whole framework but instead pick what the project requires.

So, I imported them like this so that we can use what we need and comment out the rest.

.template/addons/bootstrap/package.json.rb
rafayet-monon: You are right. But to make sure that insertion worked we write tests for it. As we have insertions like this all over the template 😅 I feel like we should change everywhere 😅 Do you think it's necessary? 🤔

.template/addons/bootstrap/stylesheets/bootstrap/bootstrap.scss.tt
gutakk: Make sense. You can go with your solution.

.template/addons/bootstrap/package.json.rb
gutakk: My bad, I forget about the tests. We can go with your solution as we can make sure that this will be added according to test.

.template/addons/bootstrap/application.scss.rb
andyduong1920: @rafayet-monon How about `insert_into_file 'app/assets/stylesheets/application.scss', after: /\/\/ Dependencies\n/ do` cc @gutakk 

as the `// Dependencies` is in that file already

image

.template/addons/bootstrap/stylesheets/bootstrap/bootstrap.scss.tt
```diff
@@ -0,0 +1,42 @@
...
+@import 'bootstrap/scss/functions';
```
andyduong1920, line 6: Does it work @rafayet-monon, as I remember we have to mention the `node_modules` folder here as well, correct me if I'm wrong

rafayet-monon: @andyduong1920 I created a new app from this branch It does work. In the image below you can see I used `row` and `col` from bootstrap. And the styles from bootstrap as well -

image

rafayet-monon: I think we don't need to mention the `node_modules` anymore is because we add it to the assets path in `initializer/assets.rb`

andyduong1920: Nice @rafayet-monon Thanks for the double-checking 🙌

.template/addons/bootstrap/application.scss.rb
rafayet-monon: It works 🎉 Changed in [ad2d080](https://github.com/nimblehq/rails-templates/pull/239/commits/ad2d0809ae33f68a296cc205772abb8ba00c7482)

gutakk:
@rafayet-monon Looks good to me

gutakk (approved):

.template/addons/bootstrap/application.js.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+insert_into_file 'app/javascript/packs/application.js', after: %r{import 'core-js/stable'.+\n} do
```
noor-ahmed, line 3: @rafayet-monon  is this needed to follow here https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/RegexpLiteral ?

rafayet-monon: I think it's as per the rule as there's a `/` in the regex. What do you think? @noor-ahmed

README.md
```diff
@@ -44,6 +44,8 @@ rails app:template LOCATION=https://raw.githubusercontent.com/nimblehq/rails-tem
 
 Available Addons:
 - `docker`
+- `semaphore`
+- `bootstrap`
```
byhbt, line 48: hi guys, now we have more addon, should we categorize them? 🤔 
```text
- Frontend:
- CI/CD:
```

.template/addons/bootstrap/application.js.rb
noor-ahmed: @rafayet-monon got it 🙌

noor-ahmed (approved):
LGTM 🍰

README.md
andyduong1920: I think we can't separate them via category for now because the ADDON list there is support for this example

Pasted_Image_10_14_20__13_47

If we separate it per category, we need to adjust the README to make it clear.

andyduong1920: IMO For updating the README content, we can do that on a separate PR @byhbt @rafayet-monon as it's out of this PR scope, this PR is focusing on adding Boostrap addon, what do you think?

byhbt (approved):
LGTM 💯

README.md
byhbt: yes, I meant to update the README content, anyway, we should do it another PR. That's fine now.

PR #241, opened 2020-10-07
Improve the localization concern, moving to around_action

### LINKED ISSUE #127: Using around_action to set the locale

andyduong1920:
Closes #127 

## Insight
Switch the locale by using **before_action** isn't safe, the locale doesn't set return correctly on some request randomly (I faced with that issue on a project).

Here is a similar issue: https://github.com/ruby-i18n/i18n/issues/381

And the official Rails guide also prefer to using around_action to switch the locale  https://guides.rubyonrails.org/i18n.html#managing-the-locale-across-requests (https://github.com/rails/rails/pull/34356)

 I also add the configuration for the I18n as since from Rails I18n 1.1.0, the config `config.i18n.fallbacks = true` (was generated by Rails) isn't correct anymore https://github.com/ruby-i18n/i18n/releases/tag/v1.1.0

## Proof Of Work

image


.template/spec/base/config/environments/production_spec.rb
```diff
@@ -23,4 +31,12 @@ def mailer_default_url_config
       }
     EOT
   end
+  
+  def i18n_config
+    <<~EOT
+      config.i18n.available_locales = ENV.fetch('AVAILABLE_LOCALES').split(', ')
```
byhbt, line 37: @andyduong1920 why we need a _space_ after `, ` here?

It means the `AVAILABLE_LOCALES="'en, th, se"` should be defined with space after each comma?

andyduong1920: Yeah, correct @byhbt ✅

config/application.yml.tt
```diff
@@ -6,6 +6,9 @@ default: &default
   MAILER_DEFAULT_HOST: "localhost"
   MAILER_DEFAULT_PORT: "3000"
   MAILER_SENDER: "Test <noreply@nimblehq.co>"
+  AVAILABLE_LOCALES: "en"
```
byhbt, line 9: @andyduong1920 I assume there will be a bit confusing here:
Should we add one more locale as an example: `"en, th"`?

Because if the user does not add the space then the locale will break, as the _parsing locale config_ code in https://github.com/nimblehq/rails-templates/pull/241/files#r503012228

andyduong1920: @byhbt I think that doesn't make sense to add more locale at default as this one is the template and not all projects need th locale, so instead of adding more example here, I think we can adding the comment in the https://github.com/nimblehq/rails-templates/pull/241/files#diff-da60b4e96eff2b132991226d308949e23f4ef3aad45ad59edd09cbc32cc6251eR24

byhbt: yes, that's fine also.

andyduong1920:
@byhbt Fixed your comment by adding the comment and change the split from `, ` to just `,` for less confusing

byhbt (approved):

.template/variants/api/app/controllers/concerns/localization.rb
```diff
@@ -2,12 +2,24 @@ module Localization
...
   included do
-    before_action :set_locale
+    around_action :switch_locale
 
     protected
 
-    def set_locale
-      I18n.locale = request.headers[:'Accept-Language'] || I18n.default_locale
+    def switch_locale(&action)
+      locale = extract_locale_from_param || I18n.default_locale
```
ankitkalia1195, line 10: @andyduong1920 How about making a function which checks for both `params[:locale]` and `request.headers[:'Accept-Language']`
```
  def get_locale
    return params[:locale].to_sym if I18n.locale_available?(params[:locale])
    return request.headers[::'Accept-Language'].to_sym if I18n.locale_available?(request.headers[::'Accept-Language'].)

    I18n.default_locale
  end

```

andyduong1920: I think that depends on each project instead of the template @ankitkalia1195 some projects only allow param, some only allow header, some allow both. So the template should be something simple and general first, and depend on each project, they could customize it on demand.

andyduong1920: And also, this PR is for changing the `before_action` to `around_action` so for improving the `get locale` logic here, we can move to another issue/pr so we could discuss and implement on that, we also need to add Unit test for that too, those are out of this PR scope ;)

rafayet-monon (approved):

noor-ahmed (approved):

.template/variants/api/app/controllers/concerns/localization.rb
ankitkalia1195: @andyduong1920 Then why are we removing `request.headers[:'Accept-Language']`if we not changing the logic ?

.template/variants/api/app/controllers/concerns/localization.rb
```diff
@@ -2,12 +2,24 @@ module Localization
...
+
+      I18n.with_locale(locale, &action)
+    end
+
+    def extract_locale_from_param
+      I18n.locale_available?(params[:locale]) ? params[:locale].to_sym : nil
```
ankitkalia1195, line 16: @andyduong1920 As per our compass convention, we should follow `return` syntax instead of conditionals, can we do that here too.

.template/variants/api/app/controllers/concerns/localization.rb
andyduong1920: Oh yeah, the original version was `request.headers[:'Accept-Language']` 😁  so should we keep that or just keep going with param locale right now 🤔 

IMO we can go with that but not sure we have the different logic between API and Web variant.

.template/variants/api/app/controllers/concerns/localization.rb
andyduong1920: will update that @ankitkalia1195 ;)

andyduong1920: Fixed in https://github.com/nimblehq/rails-templates/pull/241/commits/0e06c497f7d26bb382450f349771b6564ba93356

.template/variants/api/app/controllers/concerns/localization.rb
andyduong1920: Changed in https://github.com/nimblehq/rails-templates/pull/241/commits/1cb5bc7b734584b070666284eae456d00584991a

ankitkalia1195: @andyduong1920 How about using both as I suggested and giving preference to one?

.template/variants/api/app/controllers/concerns/localization.rb
```diff
@@ -2,12 +2,26 @@ module Localization
...
+
+    def extract_locale_from_header
+      return request.headers[:'Accept-Language'] if I18n.locale_available?(request.headers[:'Accept-Language'])
+      
+      nil
+    end
+
+    private
+
+    def default_url_options
```
ankitkalia1195, line 23: @andyduong1920 If we go for api version, we will not need `default_url_options` I guess

junan (approved):

.template/variants/api/app/controllers/concerns/localization.rb
andyduong1920: yeah, good point, removed that in https://github.com/nimblehq/rails-templates/pull/241/commits/39ea1f87d09352c5bc8e5f1bda9b50ce16a6f3c0 @ankitkalia1195

ankitkalia1195 (approved):

Issue #234, opened 2020-09-25
Enable public_file_server by default


gutakk:
### What happened?
I'm working on Ruby Internal Certification and everything about static files work fine (my case is css file).
But when I deployed to Heroku server, Rails app not be able to find css file.

This happened because 
```
# Disable serving static files from the `/public` folder by default since
# Apache or NGINX already handles this.
config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
```

### How to fix
Just set `RAILS_SERVE_STATIC_FILES` to `true`, and you are good to go.

olivierobert:
I don't recommend to set it by default as we don't always deploy to Heroku. Maybe that. could be part of a Heroku addon @andyduong1920 ?

andyduong1920:
@olivierobert as default we don't have Nginx setup so the app won't return the static file. It doesn't matter where do we deploy the app (Heroku or AWS or Google cloud) because the Rails itself doesn't return the static file.

So IMO, we could enable this by default, and we can turn it off in the [Nginx addon](https://github.com/nimblehq/rails-templates/pull/232) PR

PR #242, opened 2020-10-07
Allow Rails serve static files as default

andyduong1920:
## What happened
Solve https://github.com/nimblehq/rails-templates/issues/234

 
## Insight
Allow Rails server static by default as by default we don't have the Web server by default (like Nginx, Apache,..)

This configuration could be disabled on the [Nginx addon PR](https://github.com/nimblehq/rails-templates/pull/232)

## Proof Of Work
The test should be passed

### DIFF
--- a/.template/spec/base/config/environments/production_spec.rb
+++ b/.template/spec/base/config/environments/production_spec.rb
@@ -9,7 +9,7 @@ describe 'config/environments/production.rb' do
     expect(subject).to contain(mailer_default_url_config)
   end
 
-  it 'allow Rails serve static file' do
+  it 'allows Rails serve static file' do
     expect(subject).to contain('config.public_file_server.enabled = true')
   end
   
--- a/.template/spec/base/config/environments/production_spec.rb
+++ b/.template/spec/base/config/environments/production_spec.rb
@@ -9,6 +9,10 @@ describe 'config/environments/production.rb' do
     expect(subject).to contain(mailer_default_url_config)
   end
 
+  it 'allow Rails serve static file' do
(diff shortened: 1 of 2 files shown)


.template/spec/base/config/environments/production_spec.rb
```diff
@@ -9,6 +9,10 @@
     expect(subject).to contain(mailer_default_url_config)
   end
 
+  it 'allow Rails serve static file' do
+    expect(subject).to contain('config.public_file_server.enabled = true')
```
andyduong1920, line 13: Disabled it in Nginx addon PR https://github.com/nimblehq/rails-templates/pull/232/files#diff-c7e70c5ce93b28441e8a240fa5a19594R8

.template/spec/base/config/environments/production_spec.rb
```diff
@@ -9,6 +9,10 @@
     expect(subject).to contain(mailer_default_url_config)
   end
 
+  it 'allow Rails serve static file' do
```
gutakk, line 12: Should it be `allows` instead of `allow`, what do you think?

andyduong1920: Good catch @gutakk 👍

andyduong1920: Fixed in https://github.com/nimblehq/rails-templates/pull/242/commits/0d7f3767998d5694a9f0254fca1c53b3e9c5b182 @gutakk

gutakk (approved):
Looks good to me

olivierobert (approved):

Issue #205, opened 2020-08-18
Running assets:precompile only if there is at least one system test


andyduong1920:
### Why

Similar this PR https://github.com/nimblehq/rails-templates/pull/190

Adding the `.when_first_matching_example_defined(type: :system)` to `.template/variants/web/spec/support/assets.rb`

PR #243, opened 2020-10-07
Only run asset precompile on system test

andyduong1920:
## What happened
Prepare webpack assets only if there is at least one system test is executed, solve https://github.com/nimblehq/rails-templates/issues/205

## Insight
Just change to use when_first_matching_example_defined config
 
## Proof Of Work
Run normal tests (not type: :system) faster without webpack compilation
The webpack compilation should be run once before running system tests

### DIFF
--- a/.template/variants/web/spec/support/assets.rb
+++ b/.template/variants/web/spec/support/assets.rb
@@ -1,5 +1,5 @@
 RSpec.configure do |config|
-  config.before(:suite) do
+  config.when_first_matching_example_defined(type: :system) do
     # On CI we use Docker and the assets are precompiled (in the Dockerfile) already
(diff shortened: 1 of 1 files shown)


gutakk (approved):

.template/variants/web/spec/support/assets.rb
```diff
@@ -1,5 +1,5 @@
 RSpec.configure do |config|
-  config.before(:suite) do
+  config.when_first_matching_example_defined(type: :system) do
```
olivierobert, line 2: I did not know this existed 🤗 In the past we tried to use a flag + ENV var

olivierobert (approved):

Issue #124, opened 2019-10-16
Add Rubocop extensions


olivierobert:
## Issue

In order to increase our test coverage, it would be good to extended Rubocop base cops with Rails-spcific cops:

- Rubocop Rails: [Github](https://github.com/rubocop-hq/rubocop-rails) - RubyGems
- Rubocop Performance: [Github](https://github.com/rubocop-hq/rubocop-performance)
- Rubocop Rspec: [Github](https://github.com/rubocop-hq/rubocop-rspec)

## What's Next?

Add dependencies in Gemfile

PR #246, opened 2020-10-08
Adding rubocop-rspec

andyduong1920:
## What happened
Solved https://github.com/nimblehq/rails-templates/issues/124

 
## Insight
1/ Adding [rubocop-rspec](https://github.com/rubocop-hq/rubocop-rspec)

2/ Modify some Rubocop Rspec rule

- [RSpec/HookArgument](https://www.rubydoc.info/gems/rubocop-rspec/1.7.0/RuboCop/Cop/RSpec/HookArgument): Prefer to keep the `each` argument in the Rspec hooks., eg: `before(:each)` or `after(:each)`

- [RSpec/ContextWording](https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ContextWording): We are using `given ...` term in most of our test

## Proof Of Work
The rubocop-rspec is load 

Screen Shot 2020-10-08 at 10 31 35

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,6 @@
-require: rubocop-rails
-require: rubocop-rspec
+require: 
+  - rubocop-rails
+  - rubocop-rspec
 
 AllCops:
   Exclude:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,4 +1,5 @@
 require: rubocop-rails
+require: rubocop-rspec
 
 AllCops:
   Exclude:
@@ -33,3 +34,9 @@ Style/FormatStringToken:
 
 Rails/FilePath:
   EnforcedStyle: arguments
+
+RSpec/HookArgument:
+  EnforcedStyle: each
+
+RSpec/ContextWording:
+  Enabled: false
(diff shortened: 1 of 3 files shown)


noor-ahmed (approved):

.rubocop.yml
```diff
@@ -33,3 +35,9 @@ Style/FormatStringToken:
 
 Rails/FilePath:
   EnforcedStyle: arguments
+
+RSpec/HookArgument:
+  EnforcedStyle: each
+
+RSpec/ContextWording:
```
gutakk, line 42: Default of `ContextWording` is [(Ref)](https://github.com/rubocop-hq/rubocop-rspec/blob/master/config/default.yml#L65-L71)
```
RSpec/ContextWording:
  Description: Checks that `context` docstring starts with an allowed prefix.
  Enabled: true
  Prefixes:
  - when
  - with
  - without
```

As our convention for context is `context block descriptions should always start with ‘when’ or ‘given’`
cr. Compass

Should we do something like this instead of `Enabled: false`?

```
RSpec/ContextWording:
  Prefixes:
  - when
  - given
```

andyduong1920: Thanks for the comment, I fixed in https://github.com/nimblehq/rails-templates/pull/246/commits/f8c667c25f1d910968a3c731b4697074165d50c5 @gutakk

gutakk (approved):

olivierobert (approved):

PR #263, opened 2020-12-22
Update Rails to the latest stable version 6.1.0

longnd:
## What happened
Upgrade Rails to v6.1 - the latest stable version

 
## Insight
Remove the version constraint for the [paranoia](https://github.com/rubysherpas/paranoia) gem: we had to explicitly set it to v2.4.2 when upgrading Rails to v6.0.1 in #116 because the previous versions don't work with Rails 6.
But for now, the latest version ([v2.4.3](https://github.com/rubysherpas/paranoia/releases/tag/2.4.3)) supports Rails 6.1.0 already so that constraint isn't necessary anymore:
Screen Shot 2020-12-22 at 16 34 44

## Proof Of Work
Screen Shot 2020-12-22 at 16 28 04

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -24,7 +24,7 @@ global_job_config:
     - name: COMPOSE_FILE
       value: docker-compose.test.yml
     - name: RAILS_VERSION
-      value: 6.0.3.2
+      value: 6.1.0
   secrets:
     - name: rails-templates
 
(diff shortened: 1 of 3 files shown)


andyduong1920 (approved):
LGTM @longnd

gutakk (approved):

Lahphim (approved):

Issue #264, opened 2021-01-07
Fix: backtrace_silencers modification does not work on the latest rails version (6.1.0)


hoangmirs:
## Issue

We made a small modification on this file https://github.com/nimblehq/rails-templates/blob/develop/config/initializers/backtrace_silencers.rb that will replace the content of `config/initializers/backtrace_silencers.rb` file.

But it seems from Rails 6.0.1, they changed that file content, so we need to update the patch to work with the new content.

New version:
```ruby
# Be sure to restart your server when you modify this file.

# You can add backtrace silencers for libraries that you're using but don't wish to see in your backtraces.
# Rails.backtrace_cleaner.add_silencer { |line| /my_noisy_library/.match?(line) }

# You can also remove all the silencers if you're trying to debug a problem that might stem from framework code
# by setting BACKTRACE=1 before calling your invocation, like "BACKTRACE=1 ./bin/rails runner 'MyClass.perform'".
Rails.backtrace_cleaner.remove_silencers! if ENV["BACKTRACE"]
```

Old version:
```ruby
# Be sure to restart your server when you modify this file.

# You can add backtrace silencers for libraries that you're using but don't wish to see in your backtraces.
# Rails.backtrace_cleaner.add_silencer { |line| line =~ /my_noisy_library/ }

# You can also remove all the silencers if you're trying to debug a problem that might stem from framework code.
# Rails.backtrace_cleaner.remove_silencers!
```

## Expected

Change the content of `config/initializers/backtrace_silencers.rb` file to work with the new version
 
## Steps to reproduce

1. Install the latest rails version
2. Create a new project from our template
3. Run template test and see the error

![Upgrade_setup-ruby_action_·_nimblehq_rails-templates_0a4381f](https://user-images.githubusercontent.com/7344405/103865098-3cf81400-50f6-11eb-850c-bd701788f91d.png)

andyduong1920:
@hoangmirs A PR to fix that would be great ❤️

PR #267, opened 2021-01-12
Fix: backtrace_silencers modification does not work on the latest rails version

hoangmirs:
## What happened
Solve https://github.com/nimblehq/rails-templates/issues/264
 
## Insight
Just change the string to be replaced

## Proof Of Work
Create a new project from the template and `config/initializers/backtrace_silencers.rb` has been changed
backtrace_silencers_rb_—_test_new


andyduong1920 (approved):
👍

Issue #257, opened 2020-10-21
Add docker-prepare instruction to README template


gutakk:
## Why
When I tried to build the image (to run the test using `docker-compose -f docker-compose.test.yml run test` command) for the first time after I create the app using our template, app need to build the image and then it stuck at step 14 with following error
```sh
Step 14/22 : COPY tmp/docker ./
ERROR: Service 'test' failed to build : COPY failed: stat /var/lib/docker/tmp/docker-builder328426514/tmp/docker: no such file or directory
```
In the first place I don't know how to solve this by myself and I have to ask @hoangmirs.

@rosle explain me that because step 14 in Dockerfile try to copy stuffs in `/tmp/docker` to `./` but the `tmp/docker` directory not actually exist.

The `docker-prepare` command will create that directory for us. For more information about `docker-prepare` command [Click](https://github.com/nimblehq/rails-templates/pull/180)

## My Opinion
1. I think we can just add this instruction to `README` template file.
2. Update `README` to use `rspec` instead of `docker-compose -f docker-compose.test.yml run test` for running the test locally.

So the `Ruby and Ruby on Rails` beginner just like me will know what they have to do.
 
## Who Benefits?
- Everyone who doesn't know about `docker-prepare` command and want to use our rails template

PR #276, opened 2021-01-15
Adding `./bin/docker-prepare` into README

andyduong1920:
## What happened

Solved https://github.com/nimblehq/rails-templates/issues/257

## Insight

Adding `./bin/docker-prepare` into README

## Proof Of Work

image

### DIFF
--- a/README.md.tt
+++ b/README.md.tt
@@ -6,6 +6,13 @@
 
 ## Project Setup
 
+### Prerequisites
+
+- Ruby version: <%= RUBY_VERSION %>
+<%- if WEB_VARIANT -%>
+- Node version: <%= NODE_VERSION %>
+<%- end -%>
+
 ### Docker
 
 * Install [Docker for Mac](https://docs.docker.com/docker-for-mac/install/)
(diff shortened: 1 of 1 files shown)


README.md.tt
```diff
@@ -61,7 +61,7 @@ it first to leverage the `cache_from` settings of Docker Compose which avoids re
 * Build the Docker image:
 
 ```sh
-docker-compose -f docker-compose.test.yml build
+./bin/docker-prepare && docker-compose -f docker-compose.test.yml build
```
gutakk, line 64: How about moving `./bin/docker-prepare` to something like `Prerequisite before run test using docker or build the docker`? because
1. This command can run only once after setup the template as it just create the directory.
2. If user run test without running the build step. It will still raise the error.

What do you think @andyduong1920 ?

gutakk (commented):
Just a comment here.

README.md.tt
andyduong1920: it's for the build step @gutakk so it could be something like `Prerequisite to build the docker image`, do you have any idea for that name?

gutakk: How about put it under the `* Build the Docker image:` section something like this
```suggestion
Before run the build command, run following command once to prepare the docker
```
```sh
./bin/docker-prepare
```

You can rearrange the instruction above.

gutakk (approved):
Looks good to me 🌈

README.md.tt
andyduong1920: it needs to run every time we build the image @gutakk

bterone (approved):

PR #277, opened 2021-01-18
Make semaphoreCI addon optional

### LINKED ISSUE #265: Removing Semaphore 2.0 setting
## Why

As we are using Github Action for now, also Nimble stopped the subscription with Semaphore so for now the Semaphore CI doesn't work so far.
 
## Who Benefits?

The developer won't be confused when we having both Semaphore and Github Action

## Expected

1/ Removing the Semaphore configuration for the template itself

2/ Making the Semaphore Addon as Optional, like https://github.com/nimblehq/rails-templates/blob/develop/template.rb#L64

3/ Update the README....

longnd:
## What happened

* Remove SemaphoreCI integration for this Rails-template itself since we're using Github action to test this Rails-template.
* Make SemaphoreCI addon optional when initializing a new project using Rails-template as we choose Github action as default CI
* Do not copy the `.semaphore.yml` - SemaphoreCI classic template - when initializing a new project
* Update README file - I do not remove the mention of SemaphoreCI since the README file mentions both SemaphoreCI & Github Action since the selected addon is up to the one who setup the project using the rails-template.

Resolve #265 
## Insight

N/A

## Proof Of Work
Tests still passed
Show us the implementation: screenshots, gif, etc.


README.md.tt
```diff
@@ -45,7 +45,7 @@ foreman start -f Procfile.dev
 
 Add the following build settings to run the tests in the Docker environment via Docker Compose (configuration in `docker-compose.test.yml`):
 
-* Configure the environment variable `BRANCH_TAG` to tag Docker images per branch:
+* Configure the environment variable `BRANCH_TAG` to tag Docker images per branch, e.g:
 
 ```sh
 export BRANCH_TAG=$SEMAPHORE_BRANCH_ID
```
andyduong1920, line 51: Should we change the `SEMAPHORE_BRANCH_ID` here @longnd

andyduong1920: Or in case you think this part is redundant as we have `semaphore or github action` setup already, then we can remove this part.

template.rb
```diff
@@ -58,9 +57,9 @@ def apply_template!(template_root)
   apply '.template/addons/docker/template.rb'
   apply '.template/addons/heroku/template.rb'
   apply '.template/addons/github/template.rb'
-  apply '.template/addons/semaphore/template.rb'
 
   # Add-ons - [Optional]
+  apply '.template/addons/semaphore/template.rb' if yes?(install_addon_prompt('SemaphoreCI'))
```
andyduong1920, line 62: Can you put a TODO here because later on, we will allow the user can choose only either `Guthub Action or Semaphore`, I think it doesn't make much sense to have both CI service for 1 project (sometimes we have but it's rare)

And with this addon, the user might not be aware that Github Action doesn't include as default so their might choose Semaphore Addon, can we somehow make it clear at this point @longnd 🤔, do you have any idea?

longnd: @andyduong1920 How about this? - notifying the user about default addons installed
Screen Shot 2021-01-18 at 11 17 50

I made the change in 01f1f4f.

README.md.tt
longnd: I made the change in b4f8911 to clarify which value should be passed to BRANCH_TAG env variable. I prefer to keep the current content of README since it mentions both Github Action & SemaphoreCI. It also clarify the setup behind the Docker-based test for CI. What do you think?

template.rb
andyduong1920: Looking good for me @longnd, but later on, if we add a new `default addon` then we need to modify the `print_default_addons_message` message manually, how about we make that automatically @longnd

Something like 
```
default_addons = {
  docker: docker,
  heroku: heroku,
  github: 'Github along with Github Action'
}
```

template.rb
```diff
@@ -120,6 +121,16 @@ def print_error_message
   EOT
 end
 
+def print_default_addons_message
```
andyduong1920, line 124: the method name sounds generic...can we choose another name, many `post_default_addons_install`, maybe

template.rb
andyduong1920: @longnd if this one is a blocker to merge this PR, then we can `comment the Semaphore addon` now and create an issue to enable Semaphore addon on another PR, what do you think @longnd

longnd: thanks for your suggestion. Updated in 7953b14.

template.rb
longnd: also fixed in 281eafb.

template.rb
andyduong1920: Nice @longnd

Makefile
```diff
@@ -1,7 +1,8 @@
+# N - in response to Would you like to add the SemaphoreCI addon?
```
andyduong1920, line 1: small thing @longnd I would like to set this one as Y because, for the test, we would like to go through all addons to make sure all of that addons work, if we ignore the Semaphore CI here, this might not work as it isn't applied during the test. What do you think?

longnd: that makes sense. Updated in 71c0d78.

Makefile
```diff
@@ -1,7 +1,8 @@
+# Y - in response to Would you like to add the SemaphoreCI addon?
 # Y - in response to Would you like to add the Nginx addon?
 # Y - in response to Would you like to add the PhraseApp addon?
 # Y - in response to Would you like to add the Devise addon?
-common_addon_prompts = Y\nY\nY\n
+common_addon_prompts = N\nY\nY\nY\n
```
andyduong1920, line 5: Should we change it to `Y` @longnd ?

longnd: definitely yes, I was too hurry when making the change 🙈. Fixed in 421d5a6

andyduong1920 (approved):
LGTM @longnd

hoangmirs (approved):

PR #280, opened 2021-01-29
Fix html content matching bug

rafayet-monon:
## What happened

Error in system tests while matching HTML content in `disable_animation.rb`

## Insight

In the previous implementation system tests were not running because of wrong matching -

image

After the change, it works as expected.

## Proof Of Work

System tests running seamlessly in our internal project -

image


olivierobert (approved):
The change makes sense. But I am not sure why this was not experienced before 🙈

andyduong1920 (approved):

PR #283, opened 2021-03-22
Add more Github workflows

longnd:
## What happened

- Add 2 additional workflows: ~~Deploy to AWS Fargate &~~ Automated code review using Danger. These workflows are extracted from CX project and also partially based on the work of [nimble-survey-web](https://github.com/nimblehq/nimble-survey-web/blob/development/.github/workflows/danger.yml) by @hoangmirs 
- Update the **Pull request template** file to indicate that we need to put the story URL in the pull request - that is something we usually miss.
- Move the README file from `.template/addons/github/.github/README.md.tt` -> `.template/addons/github/.github/workflows/README.md.tt`

## Insight
- The README inside the hidden `.github` directory was created to give the instruction on how to use the workflows, not the main README file of the project which is at `README.md.tt`. But Github will prioritize the file inside the hidden directory to shows on the Github repository's page:
https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-readmes#about-readmes
> If you put your README file in your repository's root, docs, or hidden .github directory, GitHub will recognize and automatically surface your README to repository visitors.

that was the case we saw in the CX project. So we have to either rename the file inside the `.github` directory or move it inside a sub directory. I chose the latter way. 

## Proof Of Work
The workflow to deploy to AWS Fargate used in CX project:
Screen Shot 2021-03-22 at 14 51 47

The workflow of automated code review used in CX project:
Screen Shot 2021-03-22 at 15 35 09


.template/addons/github/.github/workflows/automated_code_review.yml.tt
```diff
@@ -0,0 +1,59 @@
...
+        uses: actions/cache@v2
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-nodemodules-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-nodemodules-
+
+      - name: Yarn install
+        run: yarn
+        if: |
```
olivierobert, line 53: Shouldn't the conditional be placed above the `run` command?

.template/addons/github/.github/workflows/README.md.tt
```diff
@@ -9,10 +9,13 @@
 
 ### Secrets
 
-Setup the following secrets in order to run the workflows:
+Depends on the workflow, setup the following secrets:
```
olivierobert, line 12: I don't quite get what the instructions are 😅  Do you mean:

```suggestion
Depending on the workflow used in the project, set up the following secrets:
```

longnd: that's correct. Updated in 30ac803 🙈

.template/addons/github/.github/workflows/automated_code_review.yml.tt
longnd: Yes, fixed in d69760c.

andyduong1920:
Should we set it as an optional addon @longnd? 🤔

.template/addons/github/.github/workflows/deploy_aws_fargate.yml.tt
```diff
@@ -0,0 +1,50 @@
+name: Deploy to AWS Fargate
```
andyduong1920, line 1: I think ECS is a correct name 🤔  we are deploying to the ECS and the ECS will handle the deployment to Fargate 🤔

.template/addons/github/.github/workflows/review_code_using_danger.yml.tt
```diff
@@ -0,0 +1,57 @@
...
+          restore-keys: |
+            ${{ runner.os }}-nodemodules-
+
+      - name: Yarn install
+        if: steps.yarn-cache.outputs.cache-hit != 'true' || steps.node-modules-cache.outputs.cache-hit != 'true'
+        run: yarn
+
+      <%- end -%>
+      - name: Run Danger
+        run: bundle exec danger
```
andyduong1920, line 57: should we have the bundle install step and cache the bundle path as well 🤔

longnd: it isn't necessary, the step to [setup Ruby](https://github.com/marketplace/actions/setup-ruby-jruby-and-truffleruby#single-job) handled that already (line 21)
https://github.com/nimblehq/rails-templates/blob/df4f3b35bf54a3a8747c85cf5d97361b966c80e9/.template/addons/github/.github/workflows/review_code_using_danger.yml.tt#L18-L21
according to the document of that Github action
```
bundler-cache: true # runs 'bundle install' and caches installed gems automatically
```

.template/addons/github/.github/workflows/deploy_aws_fargate.yml.tt
longnd: Fargate is a part of ECS but AWS Fargate is still the correct name, according to Amazon 😃
https://aws.amazon.com/fargate/
Screen Shot 2021-03-24 at 13 47 45

longnd:
> Should we set it as an optional addon @longnd?

What do you mean by `optional addon`? The workflows are part of Github addon and that addon is one of the default addons 🤔
https://github.com/nimblehq/rails-templates/blob/faac5fbdbfc9e6a2d1a5b97aa52ce58933c4dbf3/template.rb#L16-L20

if you meant making the workflows optional, I think the one who use the template has the right to choose the workflows they want to use after initializing a project using our template, e.g. removing the Deploy to Heroku workflow if they want to deploy to AWS only.

.template/addons/github/.github/workflows/deploy_aws_fargate.yml.tt
```diff
@@ -0,0 +1,50 @@
...
+
+      - name: Login to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v1
+
+      - name: Build Docker image
+        run: bin/docker-prepare && docker-compose build
+
+      - name: Push to ECR
+        run: docker push $DOCKER_REGISTRY_HOST/$DOCKER_IMAGE:$BRANCH_TAG
```
hoangmirs, line 47: Since you've used `docker-compose build` to build, should we use `docker-compose push` here as well?

.template/addons/github/.github/workflows/deploy_aws_fargate.yml.tt
```diff
@@ -0,0 +1,50 @@
...
+    types:
+      - completed
+  workflow_dispatch:
+
+env:
+  DOCKER_REGISTRY_HOST: ${{ secrets.DOCKER_REGISTRY_HOST }}
+  DOCKER_IMAGE: ${{ secrets.ECR_REPO }}
+  ECS_CLUSTER: ${{ secrets.ECS_CLUSTER }}
+  ECS_SERVICE: ${{ secrets.ECS_SERVICE }}
+  BRANCH_TAG: latest
```
hoangmirs, line 18: We usually use `nimblehq/branch-tag-action@v1.2` to get `BRANCH_TAG`, should we do the same to make it consistent?

longnd: I aware of that but for the main/master branch, that `branch-tag-action` action always returns the tag as `latest` so I don't think we need to spend more time running that action. What do you think?

hoangmirs: 🤔 It's unnecessary if we only run it on main/master branch. But will we run this workflow on another branch manually? If no, it's ok to set the fixed env.
 I saw you set `workflow_dispatch` event, then it can be triggered manually. In my projects, I sometimes trigger the deployment manually on staginng to test some changes before deploying to production.

.template/addons/github/.github/workflows/deploy_aws_fargate.yml.tt
longnd: this doesn't work because [docker-compose push](https://docs.docker.com/compose/reference/push/) doesn't support pushing to a private registry unless specifying in the docker-compose file, e.g. `ECR`, but [docker push](https://docs.docker.com/engine/reference/commandline/push/#push-a-new-image-to-a-registry) does

.template/addons/github/.github/workflows/deploy_aws_fargate.yml.tt
longnd: that's a good point. But I'm concern that by using the same `latest` tag for staging (and production), we only have one image on ECR, new image with the same `latest` tag will override the previous one and it saves the storing cost. But if we use the `branch-tag-action` action to generate the image tag & then deploying any random branch to staging, new image with a different tag will be pushed & left to ECR. What do you think?

.template/addons/github/.github/workflows/deploy_aws_fargate.yml.tt
andyduong1920: but we are connecting to the ECS, not the Fargate right, base on the script in this file.

.template/addons/github/.github/workflows/review_code_using_danger.yml.tt
andyduong1920: Oh I see, good to know that @longnd

andyduong1920:
> > Should we set it as an optional addon @longnd?
> 
> What do you mean by `optional addon`? The workflows are part of Github addon and that addon is one of the default addons 🤔
> https://github.com/nimblehq/rails-templates/blob/faac5fbdbfc9e6a2d1a5b97aa52ce58933c4dbf3/template.rb#L16-L20
> 
> if you meant making the workflows optional, I think the one who use the template has the right to choose the workflows they want to use after initializing a project using our template, e.g. removing the Deploy to Heroku workflow if they want to deploy to AWS only.

I mean not every project is deployed to AWS, we should set it as optional, if we don't do that the user later will remove unnecessary files (maybe they might not aware of that too), in that case the template seems not useful anymore because it generates some unnecessary stuff and required the user delete manually.

.template/addons/github/.github/workflows/deploy_aws_fargate.yml.tt
```diff
@@ -0,0 +1,50 @@
+name: Deploy to AWS Fargate
+
+on:
+  workflow_run:
+    workflows:
+      - Test
+    branches:
+      - main
```
andyduong1920, line 8: some project prefer to keep using `master`, maybe we can adding `master` along with `main` too, what do you think @longnd

longnd: I added the master branch support in ab98c46.

.template/addons/github/.github/workflows/deploy_aws_fargate.yml.tt
longnd: In this script, we must use the ECS command to deploy because Fargate is part of ECS. This script is created to target Fargate deployment only, not the EC2 with EC2. By naming `Deploy to AWS Fargate`, it removes the confusion that this workflow supports ECS with EC2 deployment.

longnd:
> I mean not every project is deployed to AWS, we should set it as optional, if we don't do that the user later will remove unnecessary files (maybe they might not aware of that too), in that case the template seems not useful anymore because it generates some unnecessary stuff and required the user delete manually.

@andyduong1920 IMO, all of the workflows here still need customization when initializing a project, e.g. the workflow to deploy to Heroku/AWS requires duplication to become 2 workflows: `Deploy to staging` & `Deploy to production` because we always need 2 environments. So asking for the workflows the user wants to use isn't really necessary. In the end, they still need to customize the generated workflows. What do you think?

.template/addons/github/.github/workflows/deploy_aws_fargate.yml.tt
andyduong1920: Please correct me if I'm wrong but I think `aws ecs update-service --cluster $ECS_CLUSTER --service $ECS_SERVICE --force-new-deployment` also work with EC2, there is nothing relative to Fargate or EC2 from this point 🤔.

longnd: @andyduong1920  it should work but I haven't had a chance to test it against EC2 deployment on ECS :D But I'm sure it works for Fargate since we used it in Nimble CX.

For EC2 deployment, we used different way in The-API https://github.com/nimblehq/the1-api/blob/development/bin/ecs-deploy

longnd: @andyduong1920 I removed the workflow to deploy to ECS from this PR. It will be handled by #289

.template/addons/github/.github/workflows/review_code_using_danger.yml.tt
```diff
@@ -0,0 +1,57 @@
+name: Automated code review
```
ankitkalia1195, line 1: @longnd Can the file name also be automated_code_review ? 🤔

longnd: I renamed the file to `review_code` in e4ccfdd to match our [naming convention](https://nimblehq.co/compass/development/code-conventions/github-actions/#naming) that requires 
> Name workflow files following the pattern `verb_environment.yml`

.template/addons/github/.github/workflows/deploy_aws_fargate.yml.tt
longnd: @hoangmirs I removed this workflow to keep this PR for the automated code review workflow only. The workflow to deploy to ECS will be included in another PR

andyduong1920 (approved):

ankitkalia1195 (approved):

.template/addons/github/.github/workflows/README.md.tt
```diff
@@ -9,10 +9,11 @@
...
 
-Setup the following secrets in order to run the workflows:
+Depending on the workflow used in the project, set up the following secrets:
 
 - DOCKER_REGISTRY_HOST
 - DOCKER_REGISTRY_TOKEN
 - DOCKER_REGISTRY_USERNAME (Default: ${{ github.repository_owner }})
 - DOCKER_IMAGE (Default: ${{ github. repository }})
 - HEROKU_API_KEY (for deployment)
+- DANGER_GITHUB_API_TOKEN (Default: ${{ github.repository_owner }})
```
hoangmirs, line 19: Is this correct?
```suggestion
- DANGER_GITHUB_API_TOKEN (Default: ${{ github.token }})
```

.template/addons/github/.github/workflows/review_code.yml.tt
```diff
@@ -0,0 +1,57 @@
+name: Review code
+
+on:
+  pull_request:
```
hoangmirs, line 4: Should we change to the following so that this workflow can be run only when we open the PR and update the code?
```suggestion
  pull_request:
    types: [opened, reopened, synchronize]
```

.template/addons/github/.github/workflows/README.md.tt
longnd: Fixed in 19db04e 🙈

.template/addons/github/.github/workflows/review_code.yml.tt
longnd: how about using the push event instead?
```suggestion
on:
  push
```

hoangmirs: @longnd I think danger needs a pull request to post the comments, please check it. I remember we had an issue when running danger without PR in `the1-api` project, that why we change to use `pull_request` instead of `push` event

longnd: yes, that's true :/ Danger check for `pull_request` event 
https://github.com/danger/danger/blob/2d63074b6ccb503642a99aaa40269a84e7067a33/lib/danger/ci_source/github_actions.rb#L23-L25
Fixed in e76bf2f.

hoangmirs (approved):

Issue #147, opened 2020-02-07
Update backtrace_cleaner to show backtrace from engines


Tolsee:
# Introduction
As we have a convention of creating `engines`, the default rails `backtrace_cleaner` is cleaning all the backtraces from `engines` in `application_trace`. 

I think we should configure the `backtrace_cleaner` so that it's easier to track errors with a proper backtrace.

Ref: 
- https://github.com/rails/rails/blob/master/railties/lib/rails/backtrace_cleaner.rb
- https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb#L76

# Example:

**Current**
![berfore](https://user-images.githubusercontent.com/16590492/74009894-c965f800-49b6-11ea-91c7-971da7650aa7.png)

Which can be improved like below, by configuring silencer 

**After**
![after](https://user-images.githubusercontent.com/16590492/74009958-edc1d480-49b6-11ea-96d5-de5ed6bebec5.png)

**NOTE:** Though, the `application_trace`(Which we get in `Stdout`) doesn't have enough traces. Our error monitoring tools(for example Sentry) have a full trace.

PR #307, opened 2022-01-06
Upgrade to Rails 7

rosle:
## What happened

- Upgrade Rails to `7.0.1`

There are new way to manage the JS and CSS asset. But I decided to do that later. Right now, in this PR, We will stick with `Webpacker` gem as before.

## Insight

To upgrade our template to support Rails 7

|Gem|Changes|Note|
|---|---|---|
|rails|Updated 6.1.1 -> 7.0.0||
|webpacker|Updated 5.2.0 -> 5.4.3||
|sass-rails|Removed|sass-rails is removed, please check out the [new guide](https://github.com/rails/rails/blob/main/guides/source/asset_pipeline.md)|
|sassc-rails|Moved to all envs scope|sass-rails is removed, please check out the [new guide](https://github.com/rails/rails/blob/main/guides/source/asset_pipeline.md)|

The new Hotwire integrations, turbo and stimulus, are also off by default following our old template.

### 📝 Note

For the next steps, we will need to:

- Check new [jsbundling-rails](https://github.com/rails/jsbundling-rails) and [cssbundling-rails](https://github.com/rails/cssbundling-rails) to replace Webpacker
- Upgrade Ruby from `2.7.2` to `3.x`

## Proof Of Work

Screen Shot 2565-01-07 at 12 13 24


config/template.rb
```diff
@@ -8,3 +8,5 @@
 apply 'config/environments/development.rb'
 apply 'config/environments/production.rb'
 apply 'config/environments/test.rb'
+
+copy_file 'config/initializers/backtrace_silencers.rb'
```
rosle, line 12: 🗒️  This initializer has been removed from the base application. so now I updated to copy file.

malparty:
That's a very interesting piece of an update! Thanks!

I was able to run it locally and display the sample page. ✅ 

Though I go a few errors on the way, I'm not sure if this was already like this in the template:

_____________________________________

- The template `rails new -m XXX` command ended up with an error:
  
  image


- Should we add in the README.md to run `bundle install` before `rake db:setup`? :) 
  
  image


- Running `rake db:setup` recommends to run `bin/rails db:migrate` but this command generates many warnings and no "success confirmation" message. 🤔 

  image

config/template.rb
olivierobert: But why add it back? Do we make use of it in general?

Gemfile.tt
```diff
@@ -2,16 +2,19 @@ source 'https://rubygems.org'
...
 gem 'puma' # Use Puma as the app server
 gem 'mini_magick' # A ruby wrapper for ImageMagick or GraphicsMagick command line
 gem 'pagy' # A pagination gem that is very light and fast
 gem 'discard' # Soft deletes for ActiveRecord
-gem 'ffaker' # A library for generating fake data such as names, addresses, and phone numbers.
-gem 'fabrication' # Fabrication generates objects in Ruby. Fabricators are schematics for your objects, and can be created as needed anywhere in your app or specs.
 gem 'sidekiq' # background processing for Ruby
 gem 'bootsnap', require: false # Reduces boot times through caching; required in config/boot.rb
+gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby] # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
+# gem 'jbuilder' # Build JSON APIs with ease
```
olivierobert, line 14: Since we rarely (or never) use `jbuilder`, how about removing it entirely?

Gemfile.tt
```diff
@@ -2,16 +2,19 @@ source 'https://rubygems.org'
...
+# gem 'redis' # Use Redis adapter to run Action Cable in production
+# gem 'kredis' # Use Kredis to get higher-level data types in Redis
```
olivierobert, line 16: Is that added by default? It feels a bit strange to have commented out dependencies 🤔

olivierobert (commented):
The update seems very complete so just a few minor questions and concerns

.template/addons/bootstrap/package.json.rb
```diff
@@ -1,8 +1,4 @@
 # frozen_string_literal: true
 
-insert_into_file 'package.json', after: /"i18n-js":.+\n/ do
-  <<~JSON
-    "bootstrap": "4.5.2",
-    "bootstrap.native": "3.0.13",
-  JSON
-end
+run 'yarn add bootstrap@^4.5.2'
+run 'yarn add bootstrap.native@^3.0.13'
```
longnd, line 4: by using Boostrap 5 which doesn't use jQuery, bootstrap native is no longer needed.

config/template.rb
rosle: Actually, I'm not sure about that. Seems we're modifying this file to make it work with engine structure. This change is from this issue - https://github.com/nimblehq/rails-templates/issues/147

Maybe we can remove it first and keep our eyes on the error. If we still have the issue, we can add it back again. What do you think?

.template/addons/bootstrap/package.json.rb
```diff
@@ -1,8 +1,4 @@
 # frozen_string_literal: true
 
-insert_into_file 'package.json', after: /"i18n-js":.+\n/ do
-  <<~JSON
-    "bootstrap": "4.5.2",
-    "bootstrap.native": "3.0.13",
-  JSON
-end
+run 'yarn add bootstrap@^4.5.2'
```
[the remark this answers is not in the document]
rosle: Indeed Long, I'll create an issue for that. 👍

rosle:
Thank you for trying it out @malparty 🎊 👨‍💻 👩‍💻 🎊 

> * The template `rails new -m XXX` command ended up with an error:
>   image

For the warning

```
warning: parser/current is loading parser/ruby27, which recognizes2.7.5-compliant syntax, but you are running 2.7.2.
Please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
```
I've also seen the parser warning too, I tried to search for the solution but it seems safe. Let me recheck it when working on upgrading Ruby to version 3. Hope that it will get rid of that warning 👍 

For the error that is blocking you from generating the project. I've never faced it. But I guess it is from the spring version 🤔  not sure why I have

Screen Shot 2565-01-12 at 15 51 18
(I use `asdf`)

For this, I will need your help to generate the project again. I'll try removing the `spring-watcher-listen` which is blocking us from having newer version of Spring -> 4.0. (Maybe I'll comment out and add todo once they support Spring 4.0, I saw someone already created the issue there.) I'll ping you to test again when finish. 💪 

-------

> * Should we add in the README.md to run `bundle install` before `rake db:setup`? :)
>   image
> * Running `rake db:setup` recommends to run `bin/rails db:migrate` but this command generates many warnings and no "success confirmation" message. 🤔
>   image

For the `db:setup`, It perfectly run fine for me right after generating the project.

Screen Shot 2565-01-12 at 16 21 37

In one of the step in template, we can see it is running `bundle install`. 

Screen Shot 2565-01-12 at 16 24 25

I'm not sure if the problem is from when generating the project from the template, it might not run successfully. 🤔  I think I need your help to try again to see if the `bundle install` is run successfully when it is generating the project.

and for the reason the `bin/rails db:migrate` does not yield any message. I think it's because we don't any the schema file in the project yet. 💭 

-------

 🕵️  I'm dropping the command to test again here so anyone could try: 🕵️ 

```
rails new <app_name> -m https://raw.githubusercontent.com/nimblehq/rails-templates/chore/upgrade-to-rails-7/template.rb
```

Gemfile.tt
rosle: Yup I brought this from the project that is generated from default rails template 🙌 

Screen Shot 2565-01-12 at 16 47 12

I put comment because I'm not sure if we will use it or not (Same for `jbuilder`)

By putting comment on them, My idea is to help us to not forget to checkout those gems later if we want to include it or not. What do you think?

malparty:
@rosle 

ℹ️ These are my sprint versions:

image

_______________


✅  I've run the command a new time: it seems better as  this time it ran **until the end** 👍  

image

_______________

> 💡 Should we just add in the README.md to run `bundle install` before the `rake db:setup` cmd? :D 

image


✅ DB migration: Yes, you are right, no schema to be deployed at the moment! This is not an issue then :) 

_______________

So overall, my issue is now fixed!! Thanks! 🚀

malparty (approved):

config/template.rb
olivierobert: Since it might be hard to reproduce the issue (unless we try creating an engine 😅 ), let's keep it. It is likely that part of Rails did not change much in that version and there this is likely still needed.

Gemfile.tt
olivierobert: Understood that it comes by default so it can be kept commented as done here

olivierobert (approved):

longnd (approved):

Issue #309, opened 2022-01-11
Upgrade Ruby to 3.x.x


rosle:
## Why

New version for Ruby is out. We better update our Ruby to LTS version. 
 
## Who Benefits?

developers

rosle:
note: Check if this issue is resolved

```
warning: parser/current is loading parser/ruby27, which recognizes2.7.5-compliant syntax, but you are running 2.7.2.
Please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
```

https://github.com/nimblehq/rails-templates/pull/307#issuecomment-1009567503

rosle:
> note: Check if this issue is resolved
> 
> ```
> warning: parser/current is loading parser/ruby27, which recognizes2.7.5-compliant syntax, but you are running 2.7.2.
> Please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
> ```
> 
> [#307 (comment)](https://github.com/nimblehq/rails-templates/pull/307#issuecomment-1009567503)

For this issue, we need to set the parser to match a version of ruby as explained here -> https://github.com/whitequark/parser#backwards-compatibility

> Parser does not use semantic versioning. Parser versions are structured as x.y.z.t, where x.y.z indicates the most recent supported Ruby release (support for every Ruby release that is chronologically earlier is implied), and t is a monotonically increasing number.

so we need to set it to `3.0.1.x`. Done [here](https://github.com/nimblehq/rails-templates/pull/312/files#diff-20b521bd9edc65692d302cfbd7c123e69e41c67c9bcc2a2e1a779b58aa934eedR55) ✅

rosle:
Please check the issue where I can't upgrade to Ruby 3.1.0 on the PR description - https://github.com/nimblehq/rails-templates/pull/312#issue-1105663837

That needs some more work 💪

rosle:
Created a new issue https://github.com/nimblehq/rails-templates/issues/316 to Upgrade Ruby to 3.1.x

PR #312, opened 2022-01-17
Upgrade to Ruby 3.0.1

rosle:
https://github.com/nimblehq/rails-templates/issues/309

## What happened

- Upgrade Ruby from `2.7.2` to `3.0.1`
- Upgrade rspec-rails to `5.0.2`
- Pin parser version to match our ruby version to fix the [parser warnings](https://github.com/nimblehq/rails-templates/pull/307#issuecomment-1009567503)
- Add `.tool-versions` file to support [asdf](https://github.com/asdf-vm/asdf)

## Insight

I also tried upgrading to Ruby 3.1.0 but faced the issue:

> /.asdf/installs/ruby/3.1.0/lib/ruby/3.1.0/psych/visitors/to_ruby.rb:430:in `visit_Psych_Nodes_Alias': Unknown alias: default (Psych::BadAlias)

After investigation, I found it is from [figaro](https://github.com/laserlemon/figaro) gem.

```
from /.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/figaro-1.2.0/lib/figaro/application.rb:60:in `parse'
from /.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/figaro-1.2.0/lib/figaro/application.rb:56:in `block in raw_configuration'
from /.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/figaro-1.2.0/lib/figaro/application.rb:56:in `raw_configuration'
from /.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/figaro-1.2.0/lib/figaro/application.rb:64:in `global_configuration'
from /.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/figaro-1.2.0/lib/figaro/application.rb:32:in `configuration'
from /.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/figaro-1.2.0/lib/figaro/application.rb:42:in `each'
from /.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/figaro-1.2.0/lib/figaro/application.rb:36:in `load'
```

This is because on Ruby 3.1.0, it comes with Psych 4 - https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/

> Psych 4.0 changes Psych.load to use safe_load by default. You may need to use Psych 3.3.2 for migrating to this behavior. [[Bug #17866]](https://bugs.ruby-lang.org/issues/17866)

In the issue, they explained that now YML is loaded with `aliases: false` by default. So when the app tries to load Figaro's `config/application.yml`, it failed because in that file, we're using alias. To fix we need to pass `aliases: true` when [loading the config](https://github.com/laserlemon/figaro/blob/6ae998e8209a08f91a165822abecffc0a0626814/lib/figaro/application.rb#L60) just like they've done on [Rails](https://github.com/rails/rails/commit/179d0a1f474ada02e0030ac3bd062fc653765dbe). 

So in order to upgrade to Ruby 3.1, we need to have this issue fixed on figaro. I'll open the issue. But if no one responds, we will need to fork or look into other alternatives. 

## Proof Of Work

To test, 

1. Install Ruby 3.0.1
2. Install Rails 7.0.1
3. Create a project from a template

> rails new $APP_NAME -m https://raw.githubusercontent.com/nimblehq/rails-templates/chore/upgrade-to-ruby-3/template.rb

4. Start the server and it should run perfectly

Screen Shot 2565-01-18 at 16 12 36


spec/support/simplecov.rb
```diff
@@ -2,7 +2,13 @@
...
 require 'simplecov-lcov'
 
-formatters = [SimpleCov::Formatter::HTMLFormatter, SimpleCov::Formatter::JSONFormatter, SimpleCov::Formatter::LcovFormatter]
+formatters = [
+  SimpleCov::Formatter::HTMLFormatter,
+  SimpleCov::Formatter::JSONFormatter,
+  SimpleCov::Formatter::LcovFormatter
+]
+
+SimpleCov::Formatter::LcovFormatter.config.report_with_single_file = true
```
rosle, line 11: Required setup for undercover - https://github.com/grodowski/undercover#setting-up-required-lcov-reporting

.template/spec/variants/web/tool_versions_spec.rb
```diff
@@ -0,0 +1,7 @@
+describe 'Web variant - .tool-versions' do
+  subject { file('.tool-versions') }
+
+  it 'pins the nodejs version' do
```
olivierobert, line 4: I am not sure of the wording `pin` (even in the context of a test) because of the meaning it has in Rails 7.x with `importmap-rails` (see [docs](https://github.com/rails/importmap-rails/)). How about instead:

```suggestion
  it 'sets the Node.JS version' do
```

olivierobert (commented):
Just a minor comment, the rest looks good to me

longnd (approved):

Issue #310, opened 2022-01-11
Checkout new asset bundling tools for Rails 7


rosle:
## Why

Rails is moving away from Webpacker by default. We should consider these new tools:

- [importmap](https://github.com/rails/importmap-rails)
- [jsbundling-rails](https://github.com/rails/jsbundling-rails)
- [cssbundling-rails](https://github.com/rails/cssbundling-rails)
 
We need to evaluate which one we should use. Please note that we need to make it work with TypeScript and Sass which are tools we use in our convention.

- [x] Integrate with new CSS bundling tool
- [x] Integrate with new JS bundling tool

## Who Benefits?

developers

Issue #115, opened 2019-09-20
Replace SCSS-Lint by Stylelint


olivierobert:
In order to make styling more agnostic (SCSS, Sass, Les), we could check is it would be good to move to https://stylelint.io/

Here is the plugin for Pronto: https://github.com/kevinjalbert/pronto-stylelint.

rosle:
We are now moved to stylelint. But haven't set up the danger to report the feedback.

rosle:
As of now, we don't have any stylelint report on the PR yet even when we migrate from pronto to danger. I'll create a new issue to integrate danger for stylelint (Maybe in the next minor version), So we can close this one 🙇

rosle:
New Issue for integrate danger stylelint - https://github.com/nimblehq/rails-templates/issues/323. I'll now close this 🙏

PR #314, opened 2022-01-21
Integrate Rails CSS Bundling

rosle:
https://github.com/nimblehq/rails-templates/issues/310
https://github.com/nimblehq/rails-templates/issues/115

## What happened

- Use [cssbundling-rails](https://github.com/rails/cssbundling-rails)
- add `bin/dev` for starting server in development env
- Replace [scss-lint](https://github.com/sds/scss-lint) gem with [stylelint](https://stylelint.io/)
- Upgrade node version from `14` -> `16`
- Update `.gitignore` to ignore `builds` and `node_modules`

#### Integrate `cssbundling-rails`

New [build css script](https://github.com/rails/cssbundling-rails/blob/main/lib/install/bootstrap/install.rb#L13). this will be run to compile the stylesheet. It is also run automatically when we do `assets:precompile`.

See more changes here:

- https://github.com/rails/cssbundling-rails/blob/main/lib/install/install.rb
- https://github.com/rails/cssbundling-rails/blob/main/lib/install/bootstrap/install.rb

#### Add `bin/dev` command

This is a helper command to start the development server instead of typing a whole foreman command

https://github.com/rails/cssbundling-rails/blob/main/lib/install/dev

#### Remove `scss-lint`

As Ruby sass is deprecating, we better move away from [scss-lint](https://github.com/sds/scss-lint) as it [depends on Ruby sass](https://github.com/sds/scss-lint/blob/master/scss_lint.gemspec#L28). We can just use Stylelint directly. The next step would be setting up Stylelint with danger

Screen Shot 2565-01-13 at 18 48 35

## Proof Of Work

The test page is working as usual. The stylesheet is applied. ✅ 

Screen Shot 2565-01-24 at 11 45 36


README.md.tt
```diff
@@ -71,7 +71,7 @@ it first to leverage the `cache_from` settings of Docker Compose which avoids re
 - Build the Docker image:
```
olivierobert, line 71: Should we also update the instructions above to run the Rails app?

Now it says:

```
- Run the Rails app
```sh
foreman start -f Procfile.dev
```
```

While we can now recommend using `bin/dev`, right?

olivierobert (approved):
This looks great to me 👍  Just one minor comment

README.md.tt
rosle: Updated on 046a4f5 ✨

junan (approved):

PR #320, opened 2022-02-03
Set up Rubocop for the template

rosle:
https://github.com/nimblehq/rails-templates/issues/168

## What happened

- Set up Rubocop for template
- Set up the workflow to run Rubocop - `Test Template`

## Insight

Due to our current template structure, there are some customization on Bundle and Rubocop.

Current structure:
```
.
├── .template
│   ├── .rspec
│   ├── .rubocop.yml   <--- Rubocop config for template here
│   ├── .ruby-gemset
│   ├── Gemfile.       <--- Gemfile for the template here
│   ├── Gemfile.lock
│   ├── ...
│   ├── addons
│   └── variants
├── bin
│   ├── ...
│   └── template.rb
├── config
│   ├── ...
│   └── template.rb
├── spec
│   ├── ...
│   └── template.rb
└── template.rb.       <--- There are also project template outside (Base template)
```  

The problem is we would need to run `rubocop` on the root of the project. So it runs tests for all the base template in the root and addon/varient in `.template`. Hence, we need to:

1. **Bundle** - Config the Gemfile path in `.bundle/config` to point to `.template/Gemfile`. With this we will able to run `bundle install` and `bundle exec rubocop` from the root of the project
2. **Rubocop** - Point the config when running rubocop command `rubocop --config .template/.rubocop.yml`. I've added the script in `.template/bin/rubocop`.

Due to the current file structure limitation, we need to have these configs in order to run rubocop. Later we can revisit the template structure e.g. consider moving base into `.template` 🤔  

## Proof Of Work

- Able to run rubocop through `.template/bin/rubocop` at the root of the project.
Screen Shot 2565-03-17 at 14 25 05

- The `Test Template` workflow is run. 
Screen Shot 2565-03-17 at 14 22 25

### DIFF
--- a/.github/workflows/test_template.yml
+++ b/.github/workflows/test_template.yml
@@ -1,6 +1,6 @@
 name: Test Template
 
-on: pull_request
+on: push
 
 env:
   RUBY_VERSION: 3.0.1
--- a/.github/workflows/test_template.yml
+++ b/.github/workflows/test_template.yml
@@ -1,0 +1,42 @@
+name: Test Template
+
+on: pull_request
+
+env:
+  RUBY_VERSION: 3.0.1
+
+jobs:
+  test:
+    name: Test template
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Cancel previous runs
+        uses: styfle/cancel-workflow-action@0.6.0
+        with:
+          access_token: ${{ github.token }}
+
+      - name: Checkout Repository
+        uses: actions/checkout@v2.3.4
+
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ env.RUBY_VERSION }}
+          bundler-cache: true
+
+      - name: Cache gems
+        uses: actions/cache@v2
+        with:
+          path: vendor/bundle
+          key: ${{ runner.os }}-template-${{ env.RUBY_VERSION }}-${{ hashFiles('**/Gemfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-template-${{ env.RUBY_VERSION }}-
+
+      - name: Install gems
+        run: |
+          bundle config path vendor/bundle
+          bundle install --jobs 4 --retry 3
+
+      - name: Run RuboCop
+        run: bundle exec rubocop --config .template/.rubocop.yml --parallel
--- a/.template/.rubocop.yml
+++ b/.template/.rubocop.yml
@@ -1,6 +1,6 @@
 AllCops:
   NewCops: enable
-  TargetRubyVersion: 2.7
+  TargetRubyVersion: 3
   Exclude:
     - 'vendor/**/*'
 
--- a/.template/.rubocop.yml
+++ b/.template/.rubocop.yml
@@ -19,13 +19,25 @@ Metrics/AbcSize:
 Metrics/BlockLength:
   Enabled: false
 
+Metrics/CyclomaticComplexity:
+  Exclude:
+    - '../template.rb'
+
+Metrics/PerceivedComplexity:
+  Exclude:
+    - '../template.rb'
+
 Naming/FileName:
+  Enabled: false
+
+# TODO: Enable this rule later
+Style/GlobalVars:
   Enabled: false
 
 # TODO: Enable this rule later
 Style/RegexpLiteral:
   Enabled: false
 
-# TODO: Enable this rule later
-Style/GlobalVars:
-  Enabled: false
+Style/TrivialAccessors:
+  Exclude:
+    - '../template.rb'
--- a/.template/.rubocop.yml
+++ b/.template/.rubocop.yml
@@ -4,11 +4,9 @@ AllCops:
   Exclude:
     - 'vendor/**/*'
 
-Style/FrozenStringLiteralComment:
-  Enabled: false
-
-Style/Documentation:
-  Enabled: false
+Lint/NonDeterministicRequireOrder:
+  Exclude:
+    - '../spec/rails_helper.rb'
 
 Metrics/MethodLength:
   Enabled: false
@@ -28,6 +26,12 @@ Metrics/PerceivedComplexity:
     - '../template.rb'
 
 Naming/FileName:
+  Enabled: false
+
+Style/Documentation:
+  Enabled: false
+
+Style/FrozenStringLiteralComment:
   Enabled: false
 
 # TODO: Enable this rule later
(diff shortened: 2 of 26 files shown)


.template/.rubocop.yml
```diff
@@ -0,0 +1,53 @@
...
+  Enabled: false
+
+Style/FrozenStringLiteralComment:
+  Enabled: false
+
+# TODO: Enable this rule later
+Style/GlobalVars:
+  Enabled: false
+
+# TODO: Enable this rule later
```
rosle, line 46: I'll enable these rules later in another PR, so this PR is not too big. 🙏

.template/lib/template/errors.rb
```diff
@@ -1 +1,3 @@
+# rubocop:todo Style/ClassAndModuleChildren
```
rosle, line 1: I'll fix this when working on refactoring the template. Because it needs to load the template module in prior. 🙏

README.md
```diff
@@ -20,11 +20,16 @@ with building complex applications over the years.
 - Install rails `7.0.1`
 - Install node `16.13.2` (For creating web application)
 
+If running on Apple M1:
+
+- Set bundle platform to ruby `bundle config set --global force_ruby_platform true`
+- To build docker image, please make sure to set platform to AMD64 `export DOCKER_DEFAULT_PLATFORM=linux/amd64`
```
olivierobert, line 26: Do you have a reference/source for this workaround? We might need to have it on other templates too.

olivierobert (approved):
Looks good to me. I just have one question.

README.md
rosle: I have issue when building the image, Chrome is failed to installed

> #9 7.669 E: Unable to locate package google-chrome-stable
Screen Shot 2565-03-17 at 16 05 22

and seems it is because we need to supply `--platform` when doing docker build

https://medium.com/no-sliver-bullet/apple-m1-%E8%AE%93%E4%BD%A0%E7%9A%84-docker-build-%E4%B8%8D%E8%B5%B7%E4%BE%86%E4%BA%86%E5%97%8E-bbe133d6d1bb

From docker documentation:
- https://docs.docker.com/engine/reference/commandline/build/#options
- https://docs.docker.com/desktop/mac/apple-silicon/#known-issues
- There are also https://docs.docker.com/desktop/multi-arch/ But I still haven't played much on it. Still don't know about it much.

We can supply that option in many places, `--platform` when build through cli, `platform` in docker compose file or through ENV `export DOCKER_DEFAULT_PLATFORM=linux/amd64`

Now I just set that ENV in my zshrc 👍

.template/.rubocop.yml
tyrro: @rosle , here are two of my favorites cops:
- [Style/GuardClause](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/GuardClause)
- [Style/OneLineConditional](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/OneLineConditional)

Is this something we could incorporate, what do you think? Thank you!

tyrro (approved):
LGTM!

README.md
rosle: For `bundle config set --global force_ruby_platform true`, without it, the `Gemfile.lock` will generate

 ```yaml
 PLATFORMS
  arm64-darwin-21
 ```
 (While usually we have platform ruby)
 
 The project is created successfully but when running docker container, it fails (I think maybe because platform is different?). The error is from Nogokiri
 
https://nokogiri.org/tutorials/installing_nokogiri.html#how-can-i-avoid-using-a-precompiled-native-gem

🛠️  To fix: we can just add ruby to platforms
 
```yaml
PLATFORMS
  arm64-darwin-21
  ruby
```
 
I found `bundle config set --global force_ruby_platform true` is easier, so we do not have to keep editing the lock file. 👀  

But after I tried to reproduce, it does not happen anymore (maybe because we set platform 💭 ). Now `Gemfile.lock` in the container is automatically modified to

```yaml
PLATFORMS
  arm64-darwin-21
  x86_64-linux
```

I'll revisit this config again tomorrow 🤔

.template/.rubocop.yml
rosle: Noted. I'll open more PRs to fix these cops and I'll revisit those rules 👍 ✨

.template/.rubocop.yml
```diff
@@ -0,0 +1,53 @@
+AllCops:
+  NewCops: enable
+  TargetRubyVersion: 2.7
```
longnd, line 3: shouldn't we target ruby 3?

longnd (commented):
Screen Shot 2022-03-18 at 10 26 29
small tip, maybe you already know:

When linking a PR to an issue, we can use [special keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to close the issue automatically when the PR is merged, e.g.
```
close #168
```

.template/.rubocop.yml
rosle: Updated in [25f0dc4](https://github.com/nimblehq/rails-templates/pull/320/commits/25f0dc400a0c1130ecb620c0eb6db4c839f746b6) ✨

README.md
rosle: I can confirm `bundle config set --global force_ruby_platform true` is not required anymore. Maybe it's from platform setting:

On local machine:
Screen Shot 2565-03-18 at 10 41 55

Inside container, The additional platform is added: 
Screen Shot 2565-03-18 at 10 55 46

I've updated the readme on [f502949](https://github.com/nimblehq/rails-templates/pull/320/commits/f5029494ef98ef3aa4ff52acf34c2e5e0af83eb3) 👍

rosle:
> When linking a PR to an issue, we can use [special keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to close the issue automatically when the PR is merged, e.g.
> 
> ```
> close #168
> ```

Thank Long 👍  There will be a couple more PRs to fix the disabled Rubocop. I disabled those rules now to prevent the PR from getting too big. I'll do it in the last PR~ ✨

.github/workflows/test_template.yml
```diff
@@ -0,0 +1,42 @@
+name: Test Template
+
+on: push
+
+env:
+  RUBY_VERSION: 3.0.1
```
junan, line 6: How about using `RUBY_VERSION` env also here?
```suggestion
  RUBY_VERSION: ${{ env.RUBY_VERSION }}
```

junan (approved):

.template/.rubocop.yml
rosle: @tyrro I revisited these 2 cops, seems we already have it enabled 👍 🚀

PR #324, opened 2022-03-18
Set up Rubocop for the template - Enable cop rules

rosle:
Close https://github.com/nimblehq/rails-templates/issues/168

## What happened

Enable more cops disabled on https://github.com/nimblehq/rails-templates/pull/320

## Insight

- Enable [regex literal style](https://docs.rubocop.org/rubocop/cops_style.html#styleregexpliteral)
- Enable [global var style](https://docs.rubocop.org/rubocop/cops_style.html#styleglobalvars), for this we find the running container 2 times, before and after suite, to start and stop test container

## Proof Of Work

The `Test Template` workflow should be 💚  along with all the tests

### DIFF
--- a/.template/.rubocop.yml
+++ b/.template/.rubocop.yml
@@ -36,9 +36,6 @@ Naming/FileName:
 Style/Documentation:
   Enabled: false
 
-Style/FrozenStringLiteralComment:
-  Enabled: false
-
 Style/TrivialAccessors:
   Exclude:
     - '../template.rb'
--- a/.template/.rubocop.yml
+++ b/.template/.rubocop.yml
@@ -39,10 +39,6 @@ Style/Documentation:
 Style/FrozenStringLiteralComment:
   Enabled: false
 
-# TODO: Enable this rule later
-Style/GlobalVars:
-  Enabled: false
-
 Style/TrivialAccessors:
   Exclude:
     - '../template.rb'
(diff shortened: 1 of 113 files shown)


.template/.rubocop.yml
```diff
@@ -39,14 +39,6 @@ Style/Documentation:
...
   Enabled: false
 
-# TODO: Enable this rule later
-Style/GlobalVars:
-  Enabled: false
-
-# TODO: Enable this rule later
-Style/RegexpLiteral:
-  Enabled: false
-
```
andyduong1920, line 49: Hi @rosle should we also enable these 2 rules

- Style/FrozenStringLiteralComment
- Naming/FileName:

rosle: For `Style/FrozenStringLiteralComment`, we can enable 🤔  seems our project rubocop also enable that and it is always auto-fixed after the project is generated.

For `Naming/FileName`, There are errors on the template file e.g. `Gemfile.rb`, `Procfile.dev.rb`, So if we want to enable this rule, I think it is possible, but we will need to exclude those files. 

I'll update and push the changes 👍 ✨

andyduong1920 (approved):
👏

malparty (approved):
LGTM, that's a lot of frozen strings 🥶🥶🥶

olivierobert (approved):

olivierobert:
@malparty Ruby 3.0 was supposed to have this enabled by default since it improves the app's performance but it was [canceled by Matz](https://bugs.ruby-lang.org/issues/11473#note-53) due to the potential breaking change it would bring 🤷

Issue #332, opened 2022-06-09
Add ability to enable Hotwire


przbadu:
## Why

Rails now ship with Hotwire in every new rails application. It would be great to choose an option to enable it.
 
## Who Benefits?

Everyone who wants to use Hotwire for their new rails application.

PR #333, opened 2022-06-10
feat(#332): ability to enable hotwire

przbadu:
## Issue
#332 

## What happened

It prompt us to enable hotwire in newly generating rails application, if the answer is `Y` then it will add
- turbo-rails
- stimulus-rails

and enable them in the rails project.

## Proof Of Work

image

image

### DIFF
--- a/.template/addons/hotwire/Gemfile.rb
+++ b/.template/addons/hotwire/Gemfile.rb
@@ -1,0 +1,8 @@
+# Hotwire Rails
+insert_into_file 'Gemfile', after: /gem 'bcrypt'.*\n/ do
+  <<~RUBY.indent(2)
+
+    gem 'turbo-rails' # Hotwire's SPA-like page accelerator
+    gem 'stimulus-rails' # Hotwire's modest JavaScript framework
+  RUBY
+end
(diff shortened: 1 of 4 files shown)


.template/addons/hotwire/Gemfile.rb
```diff
@@ -0,0 +1,8 @@
+# Hotwire Rails
+insert_into_file 'Gemfile', after: /gem 'bcrypt'.*\n/ do
+  <<~RUBY.indent(2)
```
malparty, line 3:
```suggestion
  <<~RUBY
```

Hi @przbadu  and Thanks for your contribution! :) 🙌 

I've tested it locally, the prompt is nice :) What about removing the indent as these lines are not in a group? 

For illustration, here is the resulting indentation I had at the moment:
image


Another suggestion open for your ideas, should we add a small comment before these two lines (something like `# Hotwire`) — asking @rosle @bterone  as well to share opinions :) 

Thanks!

przbadu: @malparty 

Thank you, yes those changes make sense to me.

malparty (approved):
Looks good to me! :)

bterone (approved):
Thanks for your contribution 🙌

hanam1ni (approved):

olivierobert (approved):

PR #345, opened 2022-07-15
[#344] Fix: Test failing since the Hotwire feature

### LINKED ISSUE #344: Fix: Test failing since the Hotwire feature
## Issue

As #332 was opened from a forked branch, our tests did not run on that PR. So we missed some Rubocop issues ;-) 

image

 
## Expected

The tests shall be updated in order to fully pass.
 
## Steps to reproduce

https://github.com/nimblehq/rails-templates/runs/7352294773?check_suite_focus=true

malparty:
- Close #344 

## What happened 👀

- Add missing frozen string literals
- Run rubocop auto fix (warning it tries to add an `attr_reader` in an unwanted place!)
- Move the prompt for Hotwire with the other "Web Varient Prompts"
- Add the tests for the Hotwire add-ons

## Insight 📝

> ⚠️ When a forked branch PR is submitted, the checks will not run by default and needs to be run manually by a repository owner (after ensuring the GitHub Actions are not corrupted). This is the reason why this PR has been made, to fix issues that we were not aware on a previous external PR.

## Proof Of Work 📹

All tests now pass! ✅ 
image

It prompts for Hotwire AFTER the global prompts.
image

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -67,7 +67,6 @@ def apply_template!(template_root)
   post_default_addons_install
 
   # Add-ons - [Optional]
-  apply '.template/addons/hotwire/template.rb' if WEB_VARIANT && yes?(install_addon_prompt('Hotwire'))
   apply '.template/addons/github/template.rb' if yes?(install_addon_prompt('Github Action and Wiki'))
   apply '.template/addons/semaphore/template.rb' if yes?(install_addon_prompt('SemaphoreCI'))
   apply '.template/addons/nginx/template.rb' if yes?(install_addon_prompt('Nginx'))
--- a/template.rb
+++ b/template.rb
@@ -135,7 +135,7 @@ end
 
 # Setup the template root path
 # If the template file is the url, clone the repo to the tmp directory
-template_root = __FILE__ =~ %r{\Ahttps?://} ? remote_repository : __dir__
+template_root = __FILE__.match?(%r{\Ahttps?://}) ? remote_repository : __dir__
 use_source_path template_root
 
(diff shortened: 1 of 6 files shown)


Lahphim (approved):

hoangmirs (approved):

template.rb
```diff
@@ -135,7 +134,7 @@ def get_content_between(content, string_start, string_end)
 
 # Setup the template root path
 # If the template file is the url, clone the repo to the tmp directory
-template_root = __FILE__ =~ %r{\Ahttps?://} ? remote_repository : __dir__
+template_root = __FILE__.match?(%r{\Ahttps?://}) ? remote_repository : __dir__
```
tyrro, line 137: one question: Is `=~` faster than `match?`? or are we trying to be consistent with the other places in the repo?

tyrro (approved):

bterone (approved):

template.rb
malparty: I made the change because Rubocop suggested it :D 
But it turns out:
- `=~` returns `nil` if no match and the matching strings if match.
- `.match?` is a predicate, so it returns `true` or `false` and so is slightly more performant.
- Warning to not mix with `match` (without `?`) that returns a `MatchData` object.

> 💡  [source](https://www.rubydoc.info/gems/rubocop/0.47.1/RuboCop/Cop/Performance/RegexpMatch)

But IMO, it is more a matter of consistency across all the files: The performance gain this template generation is totally negligible compared to the other tasks involved.

PR #351, opened 2022-08-05
[#350] Fix: Danger cannot find the coverage data on Github review code workflow

### LINKED ISSUE #350: Fix: Danger cannot find the coverage data on Github review code workflow
## Issue

On [EWA project](https://github.com/onward-venture/ewa-payroll-web) we found an issue using the Github `Code Review` workflow which [Danger Simplecov](https://github.com/marcelofabri/danger-simplecov_json) and [Danger Undercover](https://github.com/nimblehq/danger-undercover) complain that it cannot find the coverage data as shown in the image below.

Screen Shot 2565-08-05 at 18 25 37

Because the Danger command was run in a separate workflow as the test which generates the coverage data.
 
## Expected

The Danger should be able to find the coverage data and add a comment with the coverage percentage
 
## Steps to reproduce

1. Generate the new app with Github action
2. Push to a repository ([my test repository](https://github.com/carryall/test-danger))
3. Make some changes in another branch and open a PR ([an example PR](https://github.com/carryall/test-danger/pull/1))
4. Wait and observe the `Code Review` workflow, it went green but with this output from Danger
Screen Shot 2565-08-05 at 18 25 30

carryall:
- Close #350

## What happened 👀

The Danger cannot find the coverage data on the Github `Review Code` workflow as it runs on a separate workflow from the `Test` workflow where the coverage data is generated. 

Screen Shot 2565-08-05 at 18 25 37

So this PR fixes the issue by

- Move the code review job into the `Test` workflow
- Upload the coverage data to the artifact after the `Unit Test` job is completed
- Download the coverage data and run the Danger command

Then after the Danger command can run there's another issue about getting the output from Danger, it seems like it was trying to [call the `message` method](https://github.com/marcelofabri/danger-simplecov_json/blob/master/lib/simplecov_json/plugin.rb#L31) from the wrong place (the full log can be found [on a public repo where I test it](https://github.com/carryall/test-danger/runs/7725162900?check_suite_focus=true#step:10:17))

Screen Shot 2565-08-08 at 19 23 37

Screen Shot 2565-08-08 at 20 29 07

So I added the workaround on `Dangerfile` by copying the [report method](https://github.com/marcelofabri/danger-simplecov_json/blob/master/lib/simplecov_json/plugin.rb#L22-L35) but using the `markup` method to output the message instead

## Insight 📝

⚠️ The [Danger Undercover](https://github.com/nimblehq/danger-undercover) is also using the `message` method so the same issue happens, it only [log the output](https://github.com/carryall/test-danger/runs/7725520485?check_suite_focus=true) but does NOT add it to comment

Screen Shot 2565-08-08 at 20 34 39

What we've tried so far on the EWA project is to use [wait-on-check action](https://github.com/lewagon/wait-on-check-action) to wait for the `Test` workflow to finish before running the `Review Code` workflow but it'll double the cost of the GitHub workflow as the `Test` and `Review Code` workflow will run parallelly until the unit test job inside `Test` workflow is done

## Proof Of Work 📹

I opened [a public repository](https://github.com/carryall/test-danger) with [a test PR](https://github.com/carryall/test-danger/pull/1) with the updated workflow and files, Danger works

Screen Shot 2565-08-08 at 20 18 39

In case the `Test` job fails, the `Review Code` job also fail
Screen Shot 2565-08-11 at 15 21 37
Screen Shot 2565-08-11 at 15 21 22
with this message
Screen Shot 2565-08-11 at 15 27 33


Dangerfile
```diff
@@ -16,7 +16,25 @@ reek.lint
 suggester.suggest
 
 # Report your Ruby app test suite code coverage in Danger.
-simplecov.report 'coverage/coverage.json'
+# simplecov.report 'coverage/coverage.json'
```
malparty, line 19:
```suggestion
```

Should we remove this comment about the old way? 🤔

.template/addons/github/.github/workflows/review_code.yml.tt
```diff
@@ -1,58 +1,79 @@
...
+    timeout-minutes: 10
+    steps:
+      - name: Wait for tests build
+        uses: lewagon/wait-on-check-action@v1.1.2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          check-name: 'Build Docker image'
+          running-workflow-name: 'Wait for test build'
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          wait-interval: 20
```
malparty, line 21: I could not find an option, but is there a way to "Cancel the whole workflow, if the `Test` workflow failed"? 

Currently, the `review_code` workflow will "fail" if the `test` workflow failed. It would be clearer to have `review_code` canceled if `test` failed. 🤔 

Tho this is an optional improvement IMO :)

malparty (changes requested):
1 question and 1 suggestion~ 😇

Dangerfile
carryall: Yes, removed in [c60a339](https://github.com/nimblehq/rails-templates/pull/351/commits/c60a339cf9c58a2ab1aca2c0c5011fd4b607cf55)

.template/addons/github/.github/workflows/review_code.yml.tt
carryall: @malparty I think we can use [allowed-conclusions param](https://github.com/lewagon/wait-on-check-action#allowed-conclusions) to allow the job to run only when the job it waits for is completed with success, you can see the demonstration on the updated PR description and for the code, I updated in [78c7421](https://github.com/nimblehq/rails-templates/pull/351/commits/78c7421579c0ce58a678720188f860b30348d4c3)

malparty: It does not solve 100% my concern — i.e. having `code_review` cancelled (or `skipped`) instead of `failed`.
But as I cannot find alternatives at the moment, we can say it is good~

malparty (approved):

malparty:
@carryall I just pushed a small fix so that the tests are passing again (see 9763ba7). 
Please `git pull` on your machine before addressing any change ;-)

malparty:
@carryall should we update it as in EWA-Payroll as in the end, we had some issues with the code coverage when not run in the same workflow as the tests...? 🤔

.template/addons/github/.github/workflows/review_code.yml.tt
```diff
@@ -1,58 +1,81 @@
...
-      - name: Cache Yarn
-        uses: actions/cache@v2
-        id: yarn-cache
+        uses: actions/setup-node@v3
         with:
-          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
-          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-yarn-
+          node-version: "16.16.0"
```
andyduong1920, line 61: should we use the `NODE_VERSION` instead of hardcode the version here @carryall 🤔

malparty (approved):
Thanks @carryall 🙌

.template/addons/github/.github/workflows/review_code.yml.tt
carryall: Updated in [6a0d027](https://github.com/nimblehq/rails-templates/pull/351/commits/6a0d027d9500d065d9b71a736d02cae512342db8)

tyrro (approved):

hoangmirs (approved):

Issue #326, opened 2022-03-18
Improve assets precompilation 


olivierobert:
## Why

The template has a workaround to precompile assets for the `production` env. Since Rails seems to require all env variable to be set, the following add a dummy value:

https://github.com/nimblehq/rails-templates/blob/0e40c04eaee851257a4c19478a15337e1ba04086/bin/docker-assets-precompile#L15-L20

However, with the new build process with `esbuild`, it seems that just setting the env variable `ASSET_ENV=production` is sufficient as per [this article](https://blog.arkency.com/dont-waste-your-time-on-assets-compilation-on-heroku/#bin_rails_assets_precompile_in_a_modern_way). So it might be time to revisit.

Also, it would be good to sync assets files to S3 as well instead of serving assets from Heroku. 
 
## Who Benefits?

Web developers who work on Ruby on Rails projects

malparty:
@hoangmirs @andyduong1920 anyone willing to work on this one? :D 
If this one was implemented, we would not have spent time debugging the `missing adapter` issue (from #191 #264 )

Thanks!! 🙇

andyduong1920:
Hi @malparty, I'm not sure this is related to the https://github.com/nimblehq/rails-templates/issues/264 issue :D 

I didn't catch up with the Rails assets tooling for a while; before starting this improvement, are we using esbuild as the default tool for modification? 

Also, will we target to sync the compiled assets to S3 on the same issue 🤔 should we separate that into two different topics 🤔

malparty:
@andyduong1920 
These are 2 different issues indeed— they can be done independently I think 🤔 (starting with the assets pipeline improvement first, then configuring S3 later)

andyduong1920:
Also @malparty FYI we might improve the `asset pipeline` but for the rake task like `bin/rails i18n:js:export` it's still required all ENV 😂 because that is the Rails way FYI

malparty:
Yep, maybe we should use the NPM `i18n` package for running the export command... It might be more performance as well 😇 but this is at risk to duplicate the configuration for exporting.

PR #355, opened 2022-08-12
[#191] Add a default database adapter for production environment

### LINKED ISSUE #191: Add a default database adapter for production environment
The rails template is missing the adapter configuration for production in database.yml file
Therefore, I got the `ActiveRecord::AdapterNotSpecified` when building Docker image on production
Screen Shot 2020-07-24 at 17 30 24

malparty:
- Close #191

## What happened 👀

 Add the DATABASE_URL variable in the docker-assets-precompile script

## Insight 📝

All insights are discussed in the issue (#191)
All credits to Hoang for finding a clean fix!!

## Proof Of Work 📹

Tested on the EWA-Payroll project (as we already have the CI set up)

| Before | After |
|---|---|
| image | image |
| image | image |

### DIFF
--- a/bin/docker-assets-precompile
+++ b/bin/docker-assets-precompile
@@ -8,6 +8,7 @@
 # And it requires the envs as we always use `ENV.fetch` to setup the variables
 #
 # Related issue: https://github.com/rails/rails/issues/32947
+# TODO: https://github.com/nimblehq/rails-templates/issues/326
 
 require 'yaml'
 
--- a/bin/docker-assets-precompile
+++ b/bin/docker-assets-precompile
@@ -16,6 +16,8 @@ rails_env = ENV.fetch('RAILS_ENV', 'production')
 if rails_env == 'production'
   env_keys = YAML.load_file('config/application.yml')['docker_build'].keys
   env_keys.each { |name| ENV[name] = 'dummy_value' }
+
+  ENV['DATABASE_URL'] = 'postgres://postgres:postgres@postgres:5432/postgres'
 end
 
 exit system('bin/rails i18n:js:export && bin/rails assets:precompile')
--- a/.github/workflows/test_variants.yml
+++ b/.github/workflows/test_variants.yml
@@ -87,7 +87,7 @@ jobs:
         run: |
           export BRANCH_TAG=${{ env.BRANCH_TAG }}-${{ matrix.variant }}
           cd $APP_NAME
-          docker compose pull test || true
+          docker-compose pull test || true
 
       - name: Build docker image
         run: |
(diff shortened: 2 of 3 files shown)


malparty:
@hoangmirs @tyrro @Lahphim @junan any comment? :D

hoangmirs (approved):
LGTM

bin/docker-assets-precompile
```diff
@@ -16,6 +16,8 @@ rails_env = ENV.fetch('RAILS_ENV', 'production')
 if rails_env == 'production'
   env_keys = YAML.load_file('config/application.yml')['docker_build'].keys
   env_keys.each { |name| ENV[name] = 'dummy_value' }
```
andyduong1920, line 18: The current implementation works but can we modify this function like

```ruby
env_keys = YAML.load_file('config/application.yml')['docker_build'].keys
whitelist_envs = [DATABASE_URL]

env_keys.each do |name|
   if ![whitelist_envs].contain?(env) else
    ENV[name] = 'dummy_value'
   end
end
```

This allow us can extends if we would need to ignore any other new ENV, what do you think?

malparty: cd3d640 ✅ 

Good idea! :)

andyduong1920: My purpose there was we don't need to re-set the `DATABASE_URL` manually and keep the original value for that.

But I just checked the `config/application.yml`['docker_build'] config; seems we don't have the default value for the `DATABASE_URL`, then the `whitelist` here doesn't bring any sense 😂

Another idea, do you think we can create a new scope in the `config/application.yml` that extends from the `docker_build` 🤔 to support precompiling the asset 🤔


```ruby
docker_precompile_asset:
  <<: *docker_build
  DATABASE_URL:  postgres://postgres:postgres@postgres:5432/postgres
```

Or we can set the DATABASE_URL directly in the https://github.com/nimblehq/rails-templates/blob/develop/config/application.yml.tt as it already mentions that is mostly used for the `bin/docker-assets-precompile`

image

malparty: IMO we need to work on a cleaner long-term fix for this workaround of `precompile asset script`. 
So I think we should not change any other file for the sake of cleaning this temporary script 🤔

malparty: There is already [an issue with that](https://github.com/nimblehq/rails-templates/issues/326) and it is not the purpose of this PR (this PR aims to fix a bug in fact ^^)

andyduong1920: Then could you remove the `whitelist_envs` solution above @malparty and put a TODO link to the https://github.com/nimblehq/rails-templates/issues/326 as we will improve that there 🤔

malparty: 60189c6

tyrro (approved):
The changes in `.github/workflows/test_variants.yml` and `Makefil` are also present in https://github.com/nimblehq/rails-templates/pull/351, maybe we can undo them from this PR?

andyduong1920 (approved):

PR #359, opened 2022-09-09
[#348] Add Makefile to build commands

### LINKED ISSUE #348: Add Makefile to build commands
## Why

- Makefile helps us to build commands easily. Other templates are using it: [gin-templates](https://github.com/nimblehq/gin-templates/blob/develop/%7B%7Bcookiecutter.app_name%7D%7D/Makefile), [elixir-templates](https://github.com/nimblehq/elixir-templates/blob/680a3649dc08762882d8c02332a1f16d2f0e95a4/priv/templates/nimble_template/Makefile)
- With Makefile, we can build some basic commands e.g. `install-dependencies`, `env-setup`, `dev`, `test` so that we don't need to remember some commands in `/bin` folder. And it will be consistent with other templates as well.
 
## Who Benefits?

Rails developers

tyrro:
- Close #348

## What happened 👀

- A Makefile is added to simplify app setup and runs.
- Rubocop MaxLineLength is updated from 120 to 130 to comply with the Compass [convention](https://nimblehq.co/compass/development/code-conventions/ruby/#convention-2)

## Insight 📝

The following commands can be run through the Makefile:

- make env/setup
- make env/teardown
- make dev
- make install_dependencies
- make codebase
- make codebase/fix

## Proof Of Work 📹

- A new template is created from this branch:

```
rails new test-rails-template -m /Users/nimble/Documents/Office/Nimble/rails-templates/template.rb
```

<details>
<summary>resulting in a new `Makefile` in the created project:</summary>
Screen Shot 2023-04-07 at 5 24 00 PM
</details>

-  `make dev` running smoothly

<details>
<summary>Output of `make dev` command:</summary>
Screen Shot 2023-04-07 at 5 20 57 PM
</details>

### DIFF
--- a/Makefile.tt
+++ b/Makefile.tt
@@ -1,6 +1,6 @@
 include .env
 
-.PHONY: dev env/setup env/teardown test test/docker codebase codebase/fix
+.PHONY: dev env/setup env/teardown codebase codebase/fix
 
 dev:
 	make install-dependencies
@@ -17,19 +17,6 @@ env/teardown:  # this command will delete data
 install-dependencies:
 	bundle install
 	yarn install
-
-test:
-	docker-compose -f docker-compose.test.yml --project-name <%= APP_NAME %>_test up -d db redis
-	bundle exec rspec $(wordlist 2,$(words $(MAKECMDGOALS)),$(MAKECMDGOALS))
-	docker-compose -f docker-compose.test.yml --project-name <%= APP_NAME %>_test down
-
-test/docker:
-	./bin/docker-prepare
-	docker-compose -f docker-compose.test.yml build
-	docker-compose -f docker-compose.test.yml run test make install-dependencies
-	docker-compose -f docker-compose.test.yml run test bin/bundle exec rake db:test:prepare
-	docker-compose -f docker-compose.test.yml run test
-	docker-compose -f docker-compose.test.yml down
 
 codebase:
 	rubocop
--- a/Makefile.tt
+++ b/Makefile.tt
@@ -1,20 +1,6 @@
 include .env
 
-.PHONY: db/migrate db/rollback db/wait-for-postgres db/seed dev env/setup env/teardown test test/docker codebase codebase/fix
-
-db/migrate:
-	make db/wait-for-postgres
-	rails db:migrate
-
-db/rollback:
-	make db/wait-for-postgres
-	rails db:rollback
-
-db/wait-for-postgres:
-	$(shell DATABASE_URL=$(DATABASE_URL) ./bin/wait-for-postgres.sh)
-
-db/seed:
-	rails db:seed
+.PHONY: dev env/setup env/teardown test test/docker codebase codebase/fix
 
 dev:
 	make install-dependencies
@@ -28,17 +14,14 @@ env/setup:
 env/teardown:  # this command will delete data
 	./bin/envteardown.sh
 
-generate-sprite:
-	./bin/svg-sprite
-
 install-dependencies:
 	bundle install
 	yarn install
 
 test:
-	docker-compose -f docker-compose.test.yml --project-name ewa-payroll-web-test up -d db redis
+	docker-compose -f docker-compose.test.yml --project-name <%= APP_NAME %>_test up -d db redis
 	bundle exec rspec $(wordlist 2,$(words $(MAKECMDGOALS)),$(MAKECMDGOALS))
-	docker-compose -f docker-compose.test.yml --project-name ewa-payroll-web-test down
+	docker-compose -f docker-compose.test.yml --project-name <%= APP_NAME %>_test down
 
 test/docker:
 	./bin/docker-prepare
--- a/Makefile.tt
+++ b/Makefile.tt
@@ -1,0 +1,46 @@
+include .env
+
+.PHONY: db/migrate db/rollback db/wait-for-postgres db/seed dev env/setup env/teardown test test/docker
+
+db/migrate:
+	make db/wait-for-postgres
+	rails db:migrate
+
+db/rollback:
+	make db/wait-for-postgres
+	rails db:rollback
+
+db/wait-for-postgres:
+	$(shell DATABASE_URL=$(DATABASE_URL) ./bin/wait-for-postgres.sh)
+
+db/seed:
+	rails db:seed
+
+dev:
+	make install-dependencies
+	make env/setup
+	./bin/dev
+
+env/setup:
+	./bin/envsetup.sh
+	rails db:prepare
+
+env/teardown:  # this command will delete data
+	./bin/envteardown.sh
+
+install-dependencies:
+	bundle install
+	yarn install
+
+test:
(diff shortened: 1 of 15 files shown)


malparty:
@tyrro do you have plan to complete this PR? If yes, may you let me know when? :) Thanks!

tyrro:
> @tyrro do you have plan to complete this PR? If yes, may you let me know when? :) Thanks!
>
It's a bit late but it's ready for review now, Thanks 😅

Makefile.tt
```diff
@@ -0,0 +1,57 @@
...
+
+env/setup:
+	./bin/envsetup.sh
+	rails db:prepare
+
+env/teardown:  # this command will delete data
+	./bin/envteardown.sh
+
+generate-sprite:
+	./bin/svg-sprite
```
malparty, line 32:
```suggestion
```

I think fresh apps do not come with svg-sprite GEM 😇

Makefile.tt
```diff
@@ -0,0 +1,57 @@
...
+db/migrate:
+	make db/wait-for-postgres
+	rails db:migrate
+
+db/rollback:
+	make db/wait-for-postgres
+	rails db:rollback
+
+db/wait-for-postgres:
+	$(shell DATABASE_URL=$(DATABASE_URL) ./bin/wait-for-postgres.sh)
```
malparty, line 14: After quite some months using the `make db/*` commands, I realized they might be misleading.
Because by abstracting the db commands, we put the developer away from using properly the rails/rake DB commands.
For example, I often want to `rake db:migrate:redo`, and with the current MakeFile, it is not possible. 

Also I do not think we want to include in the make file all [the possible commands](https://gist.github.com/sergii/6105745), so what about removing these 3 commands? 🤔

malparty (changes requested):
Just an open question :) 
I tested this branch of the template locally and it is working as expected 🚀

Makefile.tt
```diff
@@ -0,0 +1,57 @@
...
+
+generate-sprite:
+	./bin/svg-sprite
+
+install-dependencies:
+	bundle install
+	yarn install
+
+test:
+	docker-compose -f docker-compose.test.yml --project-name ewa-payroll-web-test up -d db redis
```
andyduong1920, line 39: I'm not sure we are bring this into the template, for a normal project we just need to run `rspec .....` no need to start a new `docker container for run the test and stop it`

This is not a small change so we would need to bring it as a discussion or RFC to explain why we need that, as only Flexi payroll had that till now

malparty: @andyduong1920 usually, where is the DB used for tests? Is it in the same container? 🤔 
In flexi we used this way as it was conflicting with the application running on the local machine (i.e. you could not run a test when the application was running too).

tyrro: @andyduong1920 ,  not sure that I got your point 🤔 
Running tests with the docker container is already in the README:

https://github.com/nimblehq/rails-templates/blob/68125e59e5d4ef43e4d7502e800b44b529a5a87d/README.md.tt#L107

Makefile.tt
tyrro: absolutely, deleted all db related make commands in favor of using rake tasks, including `db/seeds` as well
https://github.com/nimblehq/rails-templates/pull/359/commits/c4aeed2b779d192b298409e399e41a0d61de34d3 ✅

Makefile.tt
tyrro: https://github.com/nimblehq/rails-templates/pull/359/commits/c4aeed2b779d192b298409e399e41a0d61de34d3 ✅

Makefile.tt
andyduong1920: @malparty 

1/Usually, where is the DB used for tests?
Same DB container @malparty in all Rails or Phoenix projects; we had no problem there.

I'm a bit surprised that when I can't run a test when the application is running, as it's connecting to different `databases` in each environment.

@tyrro 

It also mentioned to run without the `Docker way` 

image


And our template seems like a setting to run the `Test without docker way` on Local development, I think most developers used that, and we had no problem so far.

And you're missing understanding that point. Our problem here is, which DB container will we use for the test

1. Normally - same DB container as dev (different database name)
2. The mew Makefile introduce a new DB container for just the dev env

The new Makefile also still run the rspec from the local machine as well (non-Docker way) different as the Readme (where everything is inside the docker) FYI @tyrro.

image

---

I'm not against these changes, but I would like to know more about that, why we would need to run the test `in a different docker db container`.

This seems like introducing a new way to run the test to the template, we should have a discussion to inform the team as well.

And the last thing, I don't think the current `template` could support this new Makefile test could run, I believe we would need to modify the Database connection for the Test env.....

--

Summary:

1. Let's discuss why we need to have this, otherwise let's not introduce it into the template
2. If the team is ok with this, let's double check and modify the existing template to make sure the new Makefile test script could run 👌

tyrro: @andyduong1920 , Thanks for the detailed reply, removed `test`-related make commands in 7da2972fdb83938b24d3b3534e6c357167bd203e

@malparty , I tested the raw `rspec` command in a newly generated project as follows(all `addons` are selected)

Screen Shot 2023-04-11 at 4 28 20 PM

the contents of the `user_spec.rb` file were:

```ruby
require 'rails_helper'

RSpec.describe User, type: :model do
  it '1+1 returns 2' do
    p 'hey'
    expect(1 + 1).to eq 2
  end
end
```

I've to check with the Flexi project again to see if we are doing anything differently.

Makefile.tt
```diff
@@ -0,0 +1,27 @@
...
+env/teardown:  # this command will delete data
+	./bin/envteardown.sh
+
+install-dependencies:
+	bundle install
+	yarn install
+
+codebase:
+	rubocop
+	yarn codebase
```
andyduong1920, line 23: do we have this `command` in the `package.json` file 🤔 - same as `codebase:fix` yet?

tyrro: yes, they are:

https://github.com/nimblehq/rails-templates/blob/feb0cbc2979c8a26a7e0c200870b407e0445bae0/.template/variants/web/package.json.rb#L31-L32

andyduong1920 (approved):

Makefile.tt
malparty: I got your point now @andyduong1920, thanks for the explanation 🙌 

I just discovered/learned now that, in other projects, `tests` are run using the same container as the one that host the development DB. 
The `same` container seems OK to me. I think I configured it to another container as I got some conflicts but that was probably due to my lack of understanding of the different environments/ports/containers/DB/... 🙈 

So I think we should stick to the previous version (i.e. keep 1 single container to host both `_dev` and `_test` databases). 

One advantage of that it that we do NOT need to bring the container `down` at the end of tests (which I found super annoying in Flexi payroll, now that I use `rspec` command most of the time as it starts faster IF your container is already up).

malparty (approved):

Makefile.tt
malparty: Oh, I see you removed the `make test` command — which resolved this thread too ;-)

PR #369, opened 2022-10-14
[#363] Fix: Remove the "Has missing safe method" check from the Danger code review

### LINKED ISSUE #363: Fix: Remove the `Has missing safe method 'authorize!'` check from the Danger code review
## Issue

Our conventions require us to use `bang!` methods that do not have safe equivalent. 
But our Danger code review bot will raise a warning each time:

image

 
## Expected

No false warnings, because they create noise in the PR process.
 
## Solution

Add in the `.reek.yml` the ignore statement:
```yml
MissingSafeMethod:
    enabled: false
```

malparty:
- Close #363 

## What happened 👀

Just disabled that check :) 

Reason: We do not always have a pair of safe() and (un)safe!() method in our code. 
Example: `authorized!` is often used in controllers as per our conventions but we do not have the safe `authorize` equivalent.

## Insight 📝

`n/a`

## Proof Of Work 📹

Tested in our internal projects, we do not have this spam anymore! 👍 

| Before | After |
|---|---|
|image|image|

### DIFF
--- a/.reek.yml
+++ b/.reek.yml
@@ -4,6 +4,8 @@
 detectors:
   IrresponsibleModule:
     enabled: false
+  MissingSafeMethod:
+    enabled: false
 
   DataClump:
     max_copies: 3


longnd (approved):
The title of the PR is a bit misleading. The check applies not only for the `authorize!` method but it will remove the check for any missing safe method in general :)

malparty:
Thanks @longnd PR title and description reviewed accordingly 🙇

andyduong1920 (approved):

PR #370, opened 2022-10-14
[#362] Fix: Rubocop prevent multi except tests

### LINKED ISSUE #362: Fix: Rubocop prevent multi except tests
## Issue

While our conventions preffer single-expectation tests, it is also clear that some integration tests should use multiple expectations in order to maintain good performances. 
 
## Expected

Rubocop should not make the codebase tests fail if we are using multiple expectations in our codebase.
 
## Steps to reproduce

1. Generate a new Rails app
2. Add a system test with 2 expects.
3. Run all the tests. The codebase tests will fail.

malparty:
- Close #362 

## What happened 👀

Add 2 Rubocop rules to allow 10 expectations per spec and 5 nested blocks.

## Insight 📝

`n/a`

## Proof Of Work 📹

Quite straightforward. Tested in our current project. 😇

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -44,3 +44,9 @@ RSpec/ContextWording:
   Prefixes:
     - when
     - given
+
+RSpec/NestedGroups:
+  Max: 5
+
+RSpec/MultipleExpectations:
+  Max: 10


longnd (approved):

.rubocop.yml
```diff
@@ -44,3 +44,9 @@ RSpec/ContextWording:
   Prefixes:
     - when
     - given
+
+RSpec/NestedGroups:
+  Max: 5
+
+RSpec/MultipleExpectations:
+  Max: 10
```
carryall, line 52: Should we also exclude the system test here? as we tend to have more `expect` per test on that part?

```suggestion
  Max: 10
  Exclude:
    - "spec/systems/**/*"
```

andyduong1920 (approved):

.rubocop.yml
```diff
@@ -44,3 +44,11 @@ RSpec/ContextWording:
...
```
olivierobert, line 52: Ten seems rather high for most unit tests 😅 It is almost like having no limits (it is like disabling the cop, really). What are the use cases for this? IMO, 3-5 expectations should be be sufficient at most.

malparty:
```suggestion
  Max: 5
```

I'm fine with 5 as well now that we have excluded the system tests from this rule 🙌 
the main uses of 10 were system tests.

PR #377, opened 2023-02-01
[gh376] Separate the linting job

### LINKED ISSUE #376: Separate the linting job from the unit test
## Why

As we could have the `fast feedback` if we have an issue in the codebase.

image

 
## Who Benefits?

Developer

andyduong1920:
- close #376 

## What happened 👀

Run the lint job in a separated job so that we can get an early feedback when lint fail.

## Insight 📝

`n/a`

## Proof Of Work 📹

image

### DIFF
--- a/.template/addons/github/.github/workflows/test.yml.tt
+++ b/.template/addons/github/.github/workflows/test.yml.tt
@@ -77,7 +77,7 @@ jobs:
   unit_tests:
     name: Unit tests
     runs-on: ubuntu-latest
-    needs: linting
+    needs: build
     timeout-minutes: 15
     steps:
       - uses: actions/checkout@v3
--- a/.template/addons/github/.github/workflows/test.yml.tt
+++ b/.template/addons/github/.github/workflows/test.yml.tt
@@ -107,7 +107,7 @@ jobs:
   system_tests:
     name: System tests
     runs-on: ubuntu-latest
-    needs: linting
+    needs: build
     timeout-minutes: 15
     env:
       TEST_RETRY: 2
--- a/.template/addons/github/.github/workflows/test.yml.tt
+++ b/.template/addons/github/.github/workflows/test.yml.tt
@@ -50,10 +50,35 @@ jobs:
       - name: Push Docker image
         run: docker-compose push test
 
+  linting:
+    name: Linting
+    runs-on: ubuntu-latest
+    needs: build
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set env BRANCH_TAG
+        uses: nimblehq/branch-tag-action@v1.2
+
+      - name: Login to Docker registry
+        uses: docker/login-action@v2
+        with:
+          registry: ${{ env.DOCKER_REGISTRY_HOST }}
+          username: ${{ env.DOCKER_REGISTRY_USERNAME }}
+          password: ${{ env.DOCKER_REGISTRY_TOKEN }}
+
+      - name: Pull Docker image
+        run: docker-compose pull test || true
+
+      - name: Run codebase test
+        run: docker-compose run test bundle exec rspec spec/codebase/codebase_spec.rb --format progress
+
   unit_tests:
     name: Unit tests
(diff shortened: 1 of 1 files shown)


.template/addons/github/.github/workflows/test.yml.tt
```diff
@@ -82,15 +107,16 @@ jobs:
...
+    timeout-minutes: 15
     steps:
       - uses: actions/checkout@v3
 
       - name: Set env BRANCH_TAG
         uses: nimblehq/branch-tag-action@v1.2
 
       - name: Login to Docker registry
-        uses: docker/login-action@v1.6.0
+        uses: docker/login-action@v2
```
malparty, line 119: Should we also replace the 2 other references to this action version? 

image

malparty: (for consistency mainly)

andyduong1920: Sure, I can do that too

andyduong1920: Done in https://github.com/nimblehq/rails-templates/pull/377/commits/8a2bc6c208118470f3423100c5bd51d935dc6526

.template/addons/github/.github/workflows/test.yml.tt
```diff
@@ -50,18 +50,43 @@ jobs:
...
+        run: docker-compose pull test || true
+
+      - name: Run codebase test
+        run: docker-compose run test bundle exec rspec spec/codebase/codebase_spec.rb --format progress
+
   unit_tests:
     name: Unit tests
     runs-on: ubuntu-latest
-    needs: build
+    needs: linting
```
hoangmirs, line 80: Should we isolate 3 jobs? As we still need to run the Unit tests even if the linting fails.
And in other projects, we also have separated lint, and test workflows.
```suggestion
    needs: build
```

andyduong1920: > Should we isolate 3 jobs? As we still need to run the Unit tests even if the linting fails.

For me, I don't think we need to run the unit test there as we have to fix the codebase issue first

> And in other projects, we also have separated lint, and test workflows.

Hmm, I'm not sure about that idea, it mean we will need to move the `build` step to another workflow 🤔 I would prefer a separate PR if you prefer that way @hoangmirs and might I know which project we have separate `linting and test` workflow @hoangmirs

malparty: That's a good question 🤔 

IMO it depends what we want to favor:
a. reduce CI/CD run time? Then we better not run tests when linting is failing (more efficient)
b. unit/system tests results at the earliest? This might save 2 or 3 minutes to get the Unit Tests results as they will start BEFORE the end of the Lint workflow. (more effective)

I think we might prefer `b` (i.e. Hoang's suggestion).

malparty:
@andyduong1920 rebase will solve the tests failing + 1 discussion still open 😇 
thanks

.template/addons/github/.github/workflows/test.yml.tt
andyduong1920: In general I think we will go with the option `a`, @hoangmirs as a Web chapter lead, could you bring this topic into the Web chapter to discuss

andyduong1920: yeah, updated for both `unit test` and `system test` @malparty @hoangmirs

malparty (approved):

.template/addons/github/.github/workflows/test.yml.tt
malparty: Good for me 👍 
If you have time to bring it into Flexi :D :D :D

andyduong1920: You would like to have that on Flexi too? Will do it in on of my PR 🤞

andyduong1920: Adjustmend Flexi on the PR 583 😄

hoangmirs (approved):

PR #381, opened 2023-02-06
[#380] Fix: esbuild 0.17.X has different interface to implement the watch argument

### LINKED ISSUE #380: Fix: esbuild 0.17.X has different interface to implement the `watch` argument
## Issue

Since a few weeks, our [CI tests are failing](https://github.com/nimblehq/rails-templates/actions/runs/4073678836/jobs/7017891745) with the message: `detail: Error: Invalid option in build() call: "watch"`.
This is due to esbuild version being moved from 0.15.* to 0.17.* and a new way to implement the watch argument: https://esbuild.github.io/api/#build

## Expected

- Tests are passing
- A freshly generated web variant project should build/watch without error
 
## Steps to reproduce

1. generate a new Web variant app from the template
2. run env setup
3. run bin/dev => See the error

malparty:
- close #380

## What happened 👀

Fix esbuild watch implementation for v0.17.*

## Insight 📝

> **Note** I'm not "locking" the package.json to a specific version of `esbuild` as I believe it is better to be aware of these errors while working on long-term improvement of the template, rather than realizing that a client project was scaffolded with outdated versions.

### Tests:

Test are passing ✅ 

### Web variant project generated from this fix:

| js.1 | web.1 | 
|---|---|
|image|image|

The console has no error, stimulus is well compiled and running as expected:

image

I also tested:
- While watch, edit the controller, the build is regenerated ✅ 
- While having stop the watch (CTR+C after `bin/dev`), it stops the watch ✅ 
- running `node app/javascript/build.js` or `yarn build` will re-build the JS as expected ✅ 

image

### DIFF
--- a/.template/variants/web/app/javascript/build.js
+++ b/.template/variants/web/app/javascript/build.js
@@ -1,13 +1,17 @@
 const watch = process.argv.slice(2).includes('--watch');
 
-require('esbuild').build({
-  entryPoints: ['app/javascript/application.js'],
-  inject: ['app/javascript/global.js'],
-  bundle: true,
-  sourcemap: true,
-  watch: watch,
-  outdir: 'app/assets/builds',
-}).catch((error) => {
-  console.error(error);
-  process.exit(1);
-});
+require('esbuild')
+  .context({
+    entryPoints: ['app/javascript/application.js'],
+    inject: ['app/javascript/global.js'],
+    bundle: true,
+    sourcemap: true,
+    outdir: 'app/assets/builds',
+  })
+  .then((ctx) => {
+    watch ? ctx.watch() : ctx.rebuild().then(() => process.exit(0));
+  })
+  .catch((error) => {
+    console.error(error);
+    process.exit(1);
+  });


.template/variants/web/app/javascript/build.js
```diff
@@ -1,13 +1,17 @@
...
+    sourcemap: true,
+    outdir: 'app/assets/builds',
+  })
+  .then((ctx) => {
+    watch ? ctx.watch() : ctx.rebuild().then(() => process.exit(0));
+  })
+  .catch((error) => {
+    console.error(error);
+    process.exit(1);
+  });
```
sanG-github, line 17:
```suggestion
context = {
    entryPoints: ['app/javascript/application.js'],
    inject: ['app/javascript/global.js'],
    bundle: true,
    sourcemap: true,
    outdir: 'app/assets/builds',
}
  
require('esbuild')
  .context(context)
  .then((ctx) => {
    watch ? ctx.watch() : ctx.rebuild().then(() => process.exit(0));
  })
  .catch((error) => {
    console.error(error);
    process.exit(1);
  });
```

sanG-github (approved):
Approve with small suggestion 😄

.template/variants/web/app/javascript/build.js
olivierobert: It would help to know which benefit you want to bring with this change 😅 It looks okay, but not sure if it brings value.

There are open source tools ([example](https://github.com/bridgetownrb/bridgetown/blob/dd004e0c788d62c0381189ade475c6b18489aa73/bridgetown-core/lib/bridgetown-core/commands/esbuild/esbuild.defaults.js.erb#L191-L215)) that have far more complex Esbuild config, but the config hash is not extracted.

malparty: We are too much into Ruby and haunted by Rubocop and Reek these days xD

rda-wtag (approved):

tyrro (approved):

PR #392, opened 2023-02-17
[gh391] Enable TEST_RETRY on Github CI

### LINKED ISSUE #391: Enable System test Retry on Github action CI
## Why

On CI, due to the resource limitation (RAM, CPU) on the system test topic. As system test uses the Chrome, our system test failing randomly, we faced with that issue before and we set the `TEST_RETRY` in the existing template but we haven't enable that on Github action CI at all 😆

The reason we set it as 0 in the `application.yml` is we don't want to retry on local.

## Who Benefits?

Developer

andyduong1920:
close #391 391

## Insight 📝

Only retry on CI.

image

## Proof Of Work 📹

On the file changes tab


sanG-github (approved):

malparty (approved):
🚀 ... 💥 

... RETRY ... 

🚀  ... ✅  ✨ 👾

PR #398, opened 2023-03-23
[#378] Feature: Convert `cancel-workflow-action` to native `concurrency`

### LINKED ISSUE #378: Use GitHub native concurrency instead of the cancel-workflow action
## Why

The [cancel-workflow action added a warning on their README](https://github.com/styfle/cancel-workflow-action), that GitHub now supports concurrency and this workflow might not be needed anymore.

Resource: https://docs.github.com/en/actions/using-jobs/using-concurrency
 
## Who Benefits?

Developers: Native support means removing 1 dependency (i.e. less risk of issues and lower maintenance cost).

sanG-github:
- Close #378

## What happened 👀

- Remove `cancel-workflow-action` in all GitHub Actions workflows
- Use `concurrency` attributes in workflows to ensure only one job is run for each workflow 

## Insight 📝

We use `${{ github.workflow }}-${{ github.ref }}` for the group, which means separating it based on the `workflow` and the `ref`.

## Proof Of Work 📹

image

### DIFF
--- a/.github/workflows/test_generated_app.yml
+++ b/.github/workflows/test_generated_app.yml
@@ -15,7 +15,7 @@ env:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+  cancel-in-progress: true
 
 jobs:
   test:
--- a/.github/workflows/test_generated_app.yml
+++ b/.github/workflows/test_generated_app.yml
@@ -13,6 +13,10 @@ env:
   NODE_VERSION: 16
   RAILS_VERSION: 7.0.1
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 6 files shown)


.github/workflows/test_generated_app.yml
```diff
@@ -13,6 +13,10 @@ env:
   NODE_VERSION: 16
   RAILS_VERSION: 7.0.1
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```
hoangmirs, line 18: Why do we need to skip it in the `main` branch? Isn't it better to cancel the test when we merge multiple PRs at the same time?
Should it just be:

```suggestion
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

longnd (approved):

malparty (approved):

PR #403, opened 2023-04-28
[gh384] Add PostCSS autoprefixer

### LINKED ISSUE #384: Use PostCSS
## Why

We have PostCSS but we haven't used it yet
 
## Who Benefits?

Developers

sanG-github:
close #384 

## What happened 👀

Provide a description of the **changes** this pull request brings to the codebase. Additionally, when the pull request is still being worked on, a checklist of the planned changes is welcome to track progress.

## Insight 📝

Describe in detail why this solution is the most appropriate, which solution you tried but did not go with, and how to test the changes. References to relevant documentation are welcome as well.

## Proof Of Work 📹

Show us the implementation: screenshots, GIFs, etc.


olivierobert (approved):

.template/spec/variants/web/package_json_spec.rb
```diff
@@ -28,6 +28,11 @@
     it 'adds the script for bundling css' do
       expect(subject['scripts']).to include('build:css')
     end
+
+    it 'add the script for bundling postcss' do
```
malparty, line 32:
```suggestion
    it 'adds the script for bundling postcss' do
```

malparty (approved):
Tested on a local project ✅

suho (approved):

malparty:
@sanG-github can you rebase so that test can pass before we merge this PR? 🙇

andyduong1920:
> Thanks for the suggestion @andyduong1920, I really appreciate it 🙇
> 
> But after trying, I saw that we also need it in the `production` environments, so I removed the `--dev` option from it.

I don't think we need it on the production mode @sanG-github :D it's just a dev tool

malparty:
@andyduong1920 I'm not sure about that: When the CI/CD builds the application in "production" mode, it needs `postcss`, right? If you have the `--dev` options, it means the `postcss` package will not be installed when running the CI/CD so we cannot build the JS app (for tests and for deployment).

andyduong1920:
@malparty the build step is different with the dependency env @malparty 

- We build the output file in the Production mode doesn't mean we need the build tool is in Production dependency 

1/ We set the postcss in the devDependency
2/ We install all dependency on CI/CD - with the NODE_ENV=development -> This mean we could install all necessary tool
3/ We build the output file in the Production mode by setting the NODE_ENV=production and build


This is also similar when you build the production output file from your local .... the build file doesn't care about dependency, the bundler will figure out the needed dependency and put that into the final Production JS files build.

andyduong1920:
We could go like this for now, but better to add a TODO, so we could move the postCSS back to the devDependency cc @sanG-github

PR #405, opened 2023-05-12
[gh385] Minify the CSS and JS for production build

### LINKED ISSUE #385: Minify the CSS and JS for Production build
## Why

We haven't have it yet
 
## Who Benefits?

Developers

sanG-github:
close #385 

## What happened 👀

We customize the build process between development and `production`, as we want the `production` build to be as minimal as possible.

## Insight 📝

With JS, we config the `esbuild` configuration:
- Add `minify` option to be true in the case we don't have the `--watch` option (production)

With CSS, we separate two scripts:
- `build:css` for the production that included the `'--style=compressed'` option.
- `build:css-dev` for the `development` environment, without the compressed option.

## Proof Of Work 📹

| Minify JS                                                                                                                                                        	| Minify CSS                                                                                                                                                       	|
|------------------------------------------------------------------------------------------------------------------------------------------------------------------	|------------------------------------------------------------------------------------------------------------------------------------------------------------------	|
| Screenshot 2023-05-05 at 10 46 04 	| Screenshot 2023-05-05 at 10 52 34 	|

Generated package.json

image

Generated Procfile.dev

image

### DIFF
--- a/.template/variants/web/app/javascript/build.js
+++ b/.template/variants/web/app/javascript/build.js
@@ -6,6 +6,7 @@ require('esbuild')
     inject: ['app/javascript/global.js'],
     bundle: true,
     sourcemap: true,
+    minify: !watch,
     outdir: 'app/assets/builds',
   })
   .then((ctx) => {
--- a/.template/variants/web/package.json.rb
+++ b/.template/variants/web/package.json.rb
@@ -41,22 +41,22 @@ run 'npm set-script codebase:fix "yarn eslint:fix && yarn stylelint:fix"'
 
 source_stylesheet = 'app/assets/stylesheets/application.scss'
 bundled_stylesheet = 'app/assets/builds/application.css'
-bundled_stylesheet_options = [
+bundled_stylesheet_base_options = [
   '--no-source-map',
   '--load-path=node_modules'
 ]
-production_bundled_stylesheet_options = [
+production_bundled_stylesheet_options = bundled_stylesheet_base_options + [
   '--style=compressed'
 ]
 
 run %(npm set-script build "node app/javascript/build.js")
 run %(
-  npm set-script build:css \
-  "sass #{source_stylesheet} #{bundled_stylesheet} #{(bundled_stylesheet_options + production_bundled_stylesheet_options).join(' ')}"
+  npm set-script build:css-production \
+  "sass #{source_stylesheet} #{bundled_stylesheet} #{(production_bundled_stylesheet_options).join(' ')}"
 )
 run %(
-  npm set-script build:css-dev \
-  "sass #{source_stylesheet} #{bundled_stylesheet} #{bundled_stylesheet_options.join(' ')}"
+  npm set-script build:css \
+  "sass #{source_stylesheet} #{bundled_stylesheet} #{bundled_stylesheet_base_options.join(' ')}"
 )
 run %(npm set-script postcss "postcss public/assets/*.css --dir public/assets --config ./")
 run %(
--- a/.template/variants/web/package.json.rb
+++ b/.template/variants/web/package.json.rb
@@ -52,7 +52,7 @@ production_bundled_stylesheet_options = [
 run %(npm set-script build "node app/javascript/build.js")
 run %(
   npm set-script build:css \
-  "sass #{source_stylesheet} #{bundled_stylesheet} #{bundled_stylesheet_options.merge(production_bundled_stylesheet_options).join(' ')}"
+  "sass #{source_stylesheet} #{bundled_stylesheet} #{(bundled_stylesheet_options + production_bundled_stylesheet_options).join(' ')}"
 )
 run %(
   npm set-script build:css-dev \
--- a/.template/variants/web/package.json.rb
+++ b/.template/variants/web/package.json.rb
@@ -45,10 +45,17 @@ bundled_stylesheet_options = [
   '--no-source-map',
   '--load-path=node_modules'
 ]
+production_bundled_stylesheet_options = [
+  '--style=compressed'
+]
 
 run %(npm set-script build "node app/javascript/build.js")
 run %(
   npm set-script build:css \
+  "sass #{source_stylesheet} #{bundled_stylesheet} #{bundled_stylesheet_options.merge(production_bundled_stylesheet_options).join(' ')}"
+)
+run %(
+  npm set-script build:css-dev \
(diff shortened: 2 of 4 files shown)


.template/variants/web/package.json.rb
```diff
@@ -37,9 +37,16 @@
   '--no-source-map',
   '--load-path=node_modules'
 ]
+production_bundled_stylesheet_options = bundled_stylesheet_options + [
```
olivierobert, line 40: We might need to rename the other variable:

```ruby
bundled_stylesheet_options = [
  '--no-source-map',
  '--load-path=node_modules'
]
```

To:

```ruby
bundled_stylesheet_base_options = [
  '--no-source-map',
  '--load-path=node_modules'
]
```

It looks more descriptive:

```suggestion
production_bundled_stylesheet_options = bundled_stylesheet_base_options + [
```

.template/variants/web/package.json.rb
```diff
@@ -37,9 +37,16 @@
...
+  '--style=compressed'
+]
 
 run %(npm set-script build "node app/javascript/build.js")
 run %(
   npm set-script build:css \
```
olivierobert, line 46: It would be best to use a prefix here and avoid it below:

```
npm set-script build:css-production \
```

So `production` is special, while `dev` is the base/standard mode.

sanG-github: Thanks for the suggestion! The default one should be the `development` environment.

I've updated it in 2c9789c ✅

malparty (approved):
🔥 🚀 🔥

.template/variants/web/app/javascript/build.js
```diff
@@ -6,6 +6,7 @@ require('esbuild')
     inject: ['app/javascript/global.js'],
     bundle: true,
     sourcemap: true,
```
andyduong1920, line 8: @sanG-github a small comment, we really don't use source map on Production, should we set this as 
```suggestion
    sourcemap: !watch,
```

.template/variants/web/app/javascript/build.js
```diff
@@ -6,6 +6,7 @@ require('esbuild')
...
+    minify: !watch,
```
andyduong1920, line 9: do we have any unit test for this and the `.template/variants/web/package.json.rb` bellow @sanG-github

.template/variants/web/app/javascript/build.js
olivierobert: Wd do use `sourcemap` 😅 You might not but sourcemaps as essential for debugging

olivierobert (approved):

.template/variants/web/app/javascript/build.js
sanG-github: Thank @andyduong1920, I added test for `package.json` in 963968 ✅ But not sure how to write tests for this file 🤔

.template/variants/web/app/javascript/build.js
tyrro: @olivierobert , that's interesting, we have `--no-source-map` for `sass`, for all environments
https://github.com/nimblehq/rails-templates/blob/b913280bc0fc13f95d2406212297e4679c4d3601/.template/variants/web/package.json.rb#L37

Then, I tried enabling it for a local project, and it was trying to reference the file by its path 😄 

image

I searched around and it seems like an [existing issue](https://github.com/rails/cssbundling-rails/issues/68) for cssbundling-rails

I was wondering if you ever were able to use source maps for SCSS assets in a Rails project?

tyrro (approved):

.template/variants/web/app/javascript/build.js
malparty: @tyrro Nice catch! :o Should we create a dedicated issue to investigate this?

malparty:
@sanG-github rebase needed 🙇

.template/variants/web/app/javascript/build.js
andyduong1920: @olivierobert but we don't debugging on Production 🤔

.template/variants/web/app/javascript/build.js
andyduong1920: yeah that it ok @sanG-github

.template/variants/web/app/javascript/build.js
andyduong1920: ok, after checking this article https://css-tricks.com/should-i-use-source-maps-in-production/ it's better to have the sourcemap information in the Production build cc @tyrro @olivierobert

andyduong1920: @malparty if we have a chance, we could just remove this config on the same PR ;) 

image

andyduong1920 (approved):

.template/variants/web/app/javascript/build.js
tyrro: @andyduong1920 ,  but `source-map` doesn't work for `cssbundling-rails` (please refer to my last comment).

So if we remove `--no-source-map` option, it'll try to generate the source map, which will result in the following:

https://github.com/nimblehq/rails-templates/assets/39977805/79f48f98-5316-4ca1-9a71-96eab772de9a

which is not very helpful for debugging 😄 

@malparty , IMHO it's not worth it at this moment.
`jsbundling-rails` support `source-map` (and we are using it too btw), `css-bunding-rails` yet to support it. So, I'd keep it as it is regarding the `--no-source-map` option!

malparty:
@tyrro rebase needed and then we can merge this one 🚀

.template/variants/web/app/javascript/build.js
malparty: @andyduong1920 lets create an issue for that :)

tyrro:
@malparty , I'll tag @sanG-github so that he can do the rebasing for me 🙈 😏

.template/variants/web/app/javascript/build.js
andyduong1920: Issued created https://github.com/nimblehq/rails-templates/issues/414 @malparty @tyrro

.template/variants/web/package.json.rb
andyduong1920: The `build:css-production` won't be trigger as no one call that, the https://github.com/rails/cssbundling-rails/blob/main/lib/tasks/cssbundling/build.rake#L20 enhance the `css:build` by default, we need to go with the Sang's original solution FYI @olivierobert I will open a PR for that cc @sanG-github 

image

PR #406, opened 2023-05-19
[#404] Fix: Loaderror: cannot load such file -- sassc when generating the project from the template

### LINKED ISSUE #404: Fix: `Loaderror: cannot load such file -- sassc` when generating the project from the template
## Issue

We're facing an error when trying to generate a project from the template (without addons). It seems to require `sassc` before running the task related to Rails 7
- `rails importmap:install`
- `rails turbo:install stimulus:install`

image

It looks a bit odd as from Rails 7, we use the `cssbundling-rails` to bundle and process CSS instead of using a third party, but it works well when I try to install [sassc-rails](https://github.com/sass/sassc-rails) before moving on.

## Expected

Remove this error and try to install it with/without add-ons.

## Steps to reproduce

1. Make sure you're using Ruby version `3.0.1`
2. Run `rails new test-dev-remote -m https://raw.githubusercontent.com/nimblehq/rails-templates/main/template.rb` to generate the project from the latest develop branch.
3. Enter many times (without addons)
4. Boom 💥

sanG-github:
close #404 

## What happened 👀

We're currently failing to generate the project from our template, both API and Web. It might be caused by some dependencies added to the existing gems. 

## Insight 📝

- I tried to roll back to some old commits, but still faced the same errors -> the root cause not related to recent changes ❌ 
- I tried to add `gem 'sassc'` into the Gemfile (before `bootsnap`) -> it works ✅ 
- I tried to add `gem 'sassc'` after it (in the assets group) -> it works locally to generate the project BUT failed to run the CI ❌ 

## Proof Of Work 📹

Generate the project successfully locally.
image

### DIFF
--- a/Gemfile.tt
+++ b/Gemfile.tt
@@ -9,6 +9,7 @@ gem 'mini_magick' # A ruby wrapper for ImageMagick or GraphicsMagick command lin
 gem 'pagy' # A pagination gem that is very light and fast
 gem 'discard' # Soft deletes for ActiveRecord
 gem 'sidekiq' # background processing for Ruby
+gem 'sassc' # bootsnap dependency
 gem 'bootsnap', require: false # Reduces boot times through caching; required in config/boot.rb
 gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby] # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 # gem 'jbuilder' # Build JSON APIs with ease


Gemfile.tt
```diff
@@ -9,6 +9,7 @@ gem 'mini_magick' # A ruby wrapper for ImageMagick or GraphicsMagick command lin
 gem 'pagy' # A pagination gem that is very light and fast
 gem 'discard' # Soft deletes for ActiveRecord
 gem 'sidekiq' # background processing for Ruby
+gem 'sassc' # bootsnap dependency
```
malparty, line 12:
```suggestion
gem 'sassc', require: false # bootsnap dependency
```

Does it work if we add it as not required? 🥲

malparty: Ok, I've tested, it does not work with `require: false` 🙈

malparty (approved):

suho (approved):

PR #411, opened 2023-06-12
[gh394] Remove the Figaro gem dependency - Andy changes

andyduong1920:
close #

## What happened 👀

Provide a description of the **changes** this pull request brings to the codebase. Additionally, when the pull request is still being worked on, a checklist of the planned changes is welcome to track progress.

## Insight 📝

Describe in detail why this solution is the most appropriate, which solution you tried but did not go with, and how to test the changes. References to relevant documentation are welcome as well.

## Proof Of Work 📹

Show us the implementation: screenshots, GIFs, etc.


bin/docker-assets-precompile
```diff
@@ -0,0 +1,33 @@
...
+# Related issue: https://github.com/rails/rails/issues/32947
+# TODO: https://github.com/nimblehq/rails-templates/issues/326
+
+require 'yaml'
+
+rails_env = ENV.fetch('RAILS_ENV', 'production')
+
+if rails_env == 'production'
+  # Read the contents of the .env file
+  File.readlines('.env.development.local.example').each do |line|
```
olivierobert, line 19: If we build the Docker image, why not using the attribute `env_file`: https://docs.docker.com/compose/environment-variables/set-environment-variables/#use-the-env_file-attribute

bin/docker-assets-precompile
```diff
@@ -0,0 +1,33 @@
...
+# frozen_string_literal: true
+
+# For building production docker image
+#
+# It sets the envs inside the docker image for precompiling the assets
+# Because to precompile the assets, Rails initializes the app.
+# And it requires the envs as we always use `ENV.fetch` to setup the variables
+#
+# Related issue: https://github.com/rails/rails/issues/32947
+# TODO: https://github.com/nimblehq/rails-templates/issues/326
```
olivierobert, line 11: Technically, DHH gave a solution that works with a dummy key: https://github.com/rails/rails/issues/32947#issuecomment-1356180471

bin/docker-assets-precompile
andyduong1920: I tried the env_file but not sure why the ENV can't find there...still need time to figure out @olivierobert :D

andyduong1920:
I'm closing this PR as we might not need to remove the figaro way as the [new figaro fork](https://github.com/hlascelles/figjam) solved the issue https://github.com/nimblehq/rails-templates/pull/419

Issue #394, opened 2023-02-23
Remove the Figaro gem dependency


malparty:
## Why

In order to unlock #316 (Update Ruby 3.1).
- The Figaro gem did not receive any update since 3 years.
- Rails 7 projects now use default fallback values in the `database.yml` file (in case the ENV variable is not set). See [this example](https://github.com/zakariaf/rails-base-app/blob/main/config/database.yml)
- We already have a `.env` file for other default non-sensitive variables (the `application.yml` is actually confusing)
 
## Who Benefits?

Developers will be less confused about where the ENV default should go. Developers with the future possibility of using the next Ruby versions.

andyduong1920:
@malparty any update on this issue

malparty:
@andyduong1920 I'm struggling with tests, the contain won't build and I'm yet to understand why. Looking for debugging this when I have free time 🏃

malparty:
@andyduong1920 Update: I'm still struggling. I moved forward as the tests can execute to a much further step. But I still get errors about containers not reachable (even if previous tests where working with these containers). 

The 2 main issues:

- False negative on: https://github.com/nimblehq/rails-templates/actions/runs/5185231094/jobs/9344920937?pr=400
(when I rerun the job, it worked ><)
```
     Docker::Error::ConflictError:
       {"message":"Container 7e1f7573816916e3985b90b670453c7417575c03cc97928c9baa03ecc06b50b5 is not running"}
       
     # ./spec/base/config/environments/development_spec.rb:15:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # Excon::Error::Conflict:
     #   Expected([200, 201, 202, 203, 204, 301, 304]) <=> Actual(409 Conflict)
     #   ./spec/base/config/environments/development_spec.rb:15:in `block (2 levels) in <top (required)>'
```

- DB container not reachable on https://github.com/nimblehq/rails-templates/actions/runs/5185279454/jobs/9345169611
```
# ./spec/rails_helper.rb:13:in `<top (required)>'
# ./spec/codebase/codebase_spec.rb:3:in `require'
# ./spec/codebase/codebase_spec.rb:3:in `<top (required)>'
# ------------------
# --- Caused by: ---
# PG::ConnectionBad:
#   could not connect to server: Connection refused
#   	Is the server running on host "127.0.0.1" and accepting
#   	TCP/IP connections on port 5432?
#   could not connect to server: Cannot assign requested address
[...]

PR #419, opened 2023-07-03
[#316] Upgrade Ruby version

### LINKED ISSUE #316: Upgrade Ruby to 3.1.x
## Why

There is an attempt to upgrade it on https://github.com/nimblehq/rails-templates/pull/312

But there is an issue.

> /.asdf/installs/ruby/3.1.0/lib/ruby/3.1.0/psych/visitors/to_ruby.rb:430:in `visit_Psych_Nodes_Alias': Unknown alias: default (Psych::BadAlias)

After investigation, I found it is from [figaro](https://github.com/laserlemon/figaro) gem.

```
from /.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/figaro-1.2.0/lib/figaro/application.rb:60:in `parse'
from /.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/figaro-1.2.0/lib/figaro/application.rb:56:in `block in raw_configuration'
from /.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/figaro-1.2.0/lib/figaro/application.rb:56:in `raw_configuration'
from /.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/figaro-1.2.0/lib/figaro/application.rb:64:in `global_configuration'
from /.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/figaro-1.2.0/lib/figaro/application.rb:32:in `configuration'
from /.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/figaro-1.2.0/lib/figaro/application.rb:42:in `each'
from /.asdf/installs/ruby/3.1.0/lib/ruby/gems/3.1.0/gems/figaro-1.2.0/lib/figaro/application.rb:36:in `load'
```

This is because on Ruby 3.1.0, it comes with Psych 4 - https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/

> Psych 4.0 changes Psych.load to use safe_load by default. You may need to use Psych 3.3.2 for migrating to this behavior. [[Bug #17866]](https://bugs.ruby-lang.org/issues/17866)

In the issue, they explained that now YML is loaded with `aliases: false` by default. So when the app tries to load Figaro's `config/application.yml`, it failed because in that file, we're using alias. To fix we need to pass `aliases: true` when [loading the config](https://github.com/laserlemon/figaro/blob/6ae998e8209a08f91a165822abecffc0a0626814/lib/figaro/application.rb#L60) just like they've done on [Rails](https://github.com/rails/rails/commit/179d0a1f474ada02e0030ac3bd062fc653765dbe). 

So in order to upgrade to Ruby 3.1, we need to have this issue fixed on figaro. I'll open the issue. But if no one responds, we will need to fork or look into other alternatives. 
 
## Who Benefits?

developers

andyduong1920:
close #316

## What happened 👀

Upgrade Ruby to 3.2.2

## Insight 📝

- Replace the figaro gem with a fork version ([figjam](https://github.com/laserlemon/figaro/issues/290#issuecomment-1310934776)) that supports the Ruby 3.x ([Figaro issue](https://github.com/laserlemon/figaro/issues/289)) 
- There are other 2 issues that due to the new Ruby version
  - https://github.com/nimblehq/rails-templates/pull/419#discussion_r1250320129 
  - https://github.com/nimblehq/rails-templates/pull/419#discussion_r1250347477

## Proof Of Work 📹

The CI is passed


.template/Gemfile.lock
```diff
@@ -2,64 +2,70 @@ GEM
   remote: https://rubygems.org/
   specs:
     ast (2.4.2)
-    diff-lcs (1.4.4)
+    diff-lcs (1.5.0)
     docker-api (2.2.0)
       excon (>= 0.47.0)
       multi_json
-    excon (0.90.0)
+    excon (0.100.0)
```
andyduong1920, line 9: Need to upgrade the excon to fix the issue on https://github.com/nimblehq/rails-templates/actions/runs/5441234668/jobs/9894962870

```
An error occurred in a `before(:suite)` hook.
Failure/Error: Docker::Container.get(container_id.strip)
ArgumentError:
  The `:host` parameter should not be set for `unix://` connections.
  When supplying a `unix://` URI, it should start with `unix:/` or `unix:///`.
```

.template/addons/docker/Dockerfile.tt
```diff
@@ -38,7 +38,8 @@ RUN apt-get update -qq && \
 ADD https://dl.yarnpkg.com/debian/pubkey.gpg /tmp/yarn-pubkey.gpg
 RUN apt-key add /tmp/yarn-pubkey.gpg && rm /tmp/yarn-pubkey.gpg && \
     echo "deb http://dl.yarnpkg.com/debian/ stable main" > /etc/apt/sources.list.d/yarn.list && \
-    curl -sL https://deb.nodesource.com/setup_"$NODE_SOURCE_VERSION".x | bash - && \
+    echo 'Package: nodejs\nPin: origin deb.nodesource.com\nPin-Priority: 600' > /etc/apt/preferences.d/nodesource && \
```
andyduong1920, line 41: As Ruby-3.2.2 slim is based on the Debian-12, and Debian 12 provides a newer Node.js package 18.13.0 from the ﻿deb.debian.org repository, so to override the Node.js version, we have to pin the deb.nodesource.com first ([source](https://github.com/nodesource/distributions/issues/1579))

olivierobert:
So in the end, not moving to using https://github.com/bkeepers/dotenv?

malparty:
> * due to the ne

As this is a forked version with no guarantee for long-term support, I think it might still be preferable to move with DotEnv. In the meantime, this forked version seems preferable over the non-maintained original gem. 🤔

andyduong1920:
> So in the end, not moving to using https://github.com/bkeepers/dotenv?

The Removing Figaro issue is still there @olivierobert https://github.com/nimblehq/rails-templates/issues/394, this PR just solve the #316 they are separate issue, the dotenv replacement seems take longer than the expectation. So to solve the #316 issue, we can go with the fork figaro version first.

andyduong1920:
> > * due to the ne
> 
> As this is a forked version with no guarantee for long-term support, I think it might still be preferable to move with DotEnv. In the meantime, this forked version seems preferable over the non-maintained original gem. 🤔

Yeah I added a comment in the [Remove Figaro issue](https://github.com/nimblehq/rails-templates/issues/394#issuecomment-1617505484), we could replace Figaro with DotENV but I think that is not a small task, we have the draft PR for months already.

Btw, this PR is solving the #316 which is just upgrading the Ruby, this doesn't aims to solve the #394 ;) so there are two separate issues ;)

malparty (approved):

hoangmirs (approved):

PR #420, opened 2023-07-03
[#383] Upgrade Node and Ruby dependency versions

### LINKED ISSUE #383: Upgrade Gem versions
## Why

As we are now in 2023, we should revisit some Gem version, for example the `gem 'rspec-rails', '~> 5.0.2'`, the latest version is 6.x.x
 
## Who Benefits?

Developer

andyduong1920:
close #383

## What happened 👀

Upgrade Node and Ruby dependency versions

## Insight 📝

- Still stay as Node 16, there is a [new issue for Node 18](https://github.com/nimblehq/rails-templates/issues/421)
- I changed to use the new `npm pkg set` syntax over the `npm set-script` ([Slack message](https://nimble-co.slack.com/archives/C01GL6RBW11/p1688357829582519)) on this PR too

## Proof Of Work 📹

The CI is passed

### DIFF
--- a/Gemfile.tt
+++ b/Gemfile.tt
@@ -21,7 +21,7 @@ gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby] # Windows doe
 gem 'pundit' # Minimal authorization through OO design and pure Ruby classes
 
 # Translations
-# gem 'rails-i18n', '~> 6.0.0' # Translations for Rails
+# gem 'rails-i18n', '~> 7.0' # Translations for Rails
 
 group :development do
   gem 'foreman' # Manage Procfile-based applications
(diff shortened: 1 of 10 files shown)


Gemfile.tt
```diff
@@ -53,7 +53,7 @@ group :development, :test do
   # Code Analysis
   gem 'bullet' # help to kill N+1 queries and unused eager loading
   gem 'brakeman', require: false # A static analysis security vulnerability scanner for Ruby on Rails applications
-  gem 'parser', '3.0.1.1' # Use correct parser version to avoid parser warnings
+  gem 'parser', '3.2.2.3' # Use correct parser version to avoid parser warnings
```
olivierobert, line 56: Do we know why we need this gem?

olivierobert (approved):
Just have one small question, but the PR looks good

Gemfile.tt
andyduong1920: I didn't have time to check that as it already in the template @olivierobert there are some gem that I think we don't really use it at all but still in the Gemfile.tt, I will create a separate issue to revisit the Template Gemfile.tt 👌

andyduong1920: Issue created https://github.com/nimblehq/rails-templates/issues/426

malparty (approved):

sanG-github (approved):

PR #423, opened 2023-07-03
[#416] Remove sassc

### LINKED ISSUE #416: Remove the dependency with the gem `sassc`
## Why

The sassc gem is old and poorly updated. It is best if we can remove it before an issue occurs due to it.

> **Note** it was added in https://github.com/nimblehq/rails-templates/pull/406 following a bug that prevented using the template. Make sure to carefully test when removing the gem.

## Who Benefits?

Developers.

andyduong1920:
close #416

## What happened 👀

Removed the sassc gem

## Insight 📝

I tried on local following the `Steps to reproduce` on https://github.com/nimblehq/rails-templates/issues/404 with

- Ruby 3.2.2
- Rails 7.0.6
- Without any addon

The project could be generated without any error

image

## Proof Of Work 📹

No error on the `rails  importmap:install` or `rails  turbo:install stimulus:install` step

image

### DIFF
--- a/Gemfile.tt
+++ b/Gemfile.tt
@@ -9,7 +9,6 @@ gem 'mini_magick' # A ruby wrapper for ImageMagick or GraphicsMagick command lin
 gem 'pagy' # A pagination gem that is very light and fast
 gem 'discard' # Soft deletes for ActiveRecord
 gem 'sidekiq' # background processing for Ruby
-gem 'sassc' # bootsnap dependency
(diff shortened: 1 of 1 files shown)


malparty (approved):
FYI, I tested with an older version of Ruby and the error message is clear because now it will tell "Gemfile expect ruby 3.2.2" before running into the error we used to have 🥳 

Very nice that we can remove this unwanted dependency!!!! 🙏

olivierobert (approved):

sanG-github (approved):
Thank you!!!

PR #427, opened 2023-07-04
Correct the build:css task

andyduong1920:
## What happened 👀

Correct the `yarn task` for `build:css` as the `build:css-production` won't be trigger by any party, 

## Insight 📝

1. The [css-bundling](https://github.com/rails/cssbundling-rails/blob/main/lib/tasks/cssbundling/build.rake#L20) calls the `build:css` not `build:css-production` in the asset:precompile, see https://github.com/nimblehq/rails-templates/pull/405#discussion_r1251536103
2. For the `build:css-dev` we will use it in the `Procfile.dev`

## Proof Of Work 📹

The CI is passed

### DIFF
--- a/.template/variants/web/.gitignore.rb
+++ b/.template/variants/web/.gitignore.rb
@@ -26,5 +26,9 @@ append_to_file '.gitignore' do
 
     # Ignore Gemfile.lock file in engines.
     /engines/*/Gemfile.lock
+
+    # Ignore asset builds
+    /app/assets/builds/*
+    !/app/assets/builds/.keep
   IGNORE
 end
--- a/.template/variants/web/.gitignore.rb
+++ b/.template/variants/web/.gitignore.rb
@@ -26,9 +26,5 @@ append_to_file '.gitignore' do
 
     # Ignore Gemfile.lock file in engines.
     /engines/*/Gemfile.lock
-
-    # Ignore asset builds
-    /app/assets/builds/*
-    !/app/assets/builds/.keep
   IGNORE
 end
--- a/.template/variants/web/app/javascript/build.js
+++ b/.template/variants/web/app/javascript/build.js
@@ -2,7 +2,10 @@ const watch = process.argv.slice(2).includes('--watch');
 
 require('esbuild')
   .context({
-    entryPoints: ['app/javascript/application.js'],
+    entryPoints: [
+      { in: 'app/javascript/application.js', out: 'application.js' },
+      // { in: 'engines/[engine_name]/app/javascript/[engine_name]/application.js', out: '[engine_name]/application.js' },
+    ],
     inject: ['app/javascript/global.js'],
     bundle: true,
     sourcemap: true,
--- a/.template/variants/web/app/javascript/build.js
+++ b/.template/variants/web/app/javascript/build.js
@@ -3,8 +3,8 @@ const watch = process.argv.slice(2).includes('--watch');
 require('esbuild')
   .context({
     entryPoints: [
-      { in: 'app/javascript/application.js', out: 'application.js' },
-      // { in: 'engines/[engine_name]/app/javascript/[engine_name]/application.js', out: '[engine_name]/application.js' },
+      { in: 'app/javascript/application.js', out: 'application' },
+      // { in: 'engines/[engine_name]/app/javascript/[engine_name]/application.js', out: '[engine_name]/application' },
     ],
     inject: ['app/javascript/global.js'],
     bundle: true,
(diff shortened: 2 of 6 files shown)


.template/variants/web/app/javascript/build.js
```diff
@@ -2,7 +2,10 @@ const watch = process.argv.slice(2).includes('--watch');
 
 require('esbuild')
   .context({
-    entryPoints: ['app/javascript/application.js'],
+    entryPoints: [
+      { in: 'app/javascript/application.js', out: 'application.js' },
```
andyduong1920, line 6: also take this changes to add this placeholder for engine JS file, so the future developer no need to spend time to figure out this.

I have tested it in our project

image

.template/variants/web/.gitignore.rb
```diff
@@ -26,5 +26,9 @@
 
     # Ignore Gemfile.lock file in engines.
     /engines/*/Gemfile.lock
+
+    # Ignore asset builds
+    /app/assets/builds/*
+    !/app/assets/builds/.keep
```
tyrro, line 32: @andyduong1920 , it's already there at the beginning of this file, do we need to add it again? 🤔 

https://github.com/nimblehq/rails-templates/blob/1a3bbf565ea1702e91f43c14e14cf44b3fb80475/.template/variants/web/.gitignore.rb#L9-L11

andyduong1920: my bad @tyrro removed it 🤞

.template/variants/web/app/javascript/build.js
```diff
@@ -2,12 +2,18 @@ const watch = process.argv.slice(2).includes('--watch');
...
+      // { in: 'engines/[engine_name]/app/javascript/[engine_name]/application.js', out: '[engine_name]/application.js' },
+    ],
     inject: ['app/javascript/global.js'],
     bundle: true,
     sourcemap: true,
     minify: !watch,
     outdir: 'app/assets/builds',
+    alias: {
+      'Core': './app/javascript/',
```
andyduong1920, line 15: I also take this chance to add the Alias here, so the future developer no need to research for it, let me know if we would like to remove it from the template 👌 

The alias will allow us to refer to the core js path from the engine JS file with the `Core/components/....` like the image bellow 

image

malparty: Should we rename it `MainApp` or `Root`?

`Core` means "central", but in the case of a simple presentation layer App that relies on `gems` or `engines` for the core, the `main_app` might not be the `core` of the system 💭

malparty (approved):
Nice changes!!!

.template/variants/web/app/javascript/build.js
tyrro: I am not sure about the usage of [`alias`](https://esbuild.github.io/api/#alias) here.  According to the doc:

> This feature lets you substitute one package for another when bundling. One use case for this feature is replacing a node-only package with a browser-friendly package in third-party code that you don't control.

I think what we're trying to achieve here belongs to the `path` of [`jsconfig.json`](https://stackoverflow.com/questions/68675994/what-is-jsconfig-json) inside the Rails project and [esbuild supports jsconfig.json path as well](https://github.com/evanw/esbuild/issues/38).

An example of `jsconfig.json` for this case could be:

```json
{
  "compilerOptions": {
    "baseUrl": "./app/javascript",
    "paths": {
      "Core/helpers/*": ["helpers/*"]
    }
  }
}
```

Then we could import the helpers exactly how we're doing right now!

malparty: Nice catch Rajib 🙇

andyduong1920: I'm not aware about that @tyrro, let me check that

andyduong1920: Hi @tyrro I tried that and seems it doesn't work FYI

image

I will remove the changes in this PR FYI

andyduong1920: @tyrro I removed the config in this [commit](https://github.com/nimblehq/rails-templates/pull/427/commits/bec259c6835c8f0f09a53a547e1919724a32698c), cc @malparty if you think this is needed for the template, please create the new issue for that 👌

malparty: Not a must-have, it is quite an opinionated topic from what I remember ^^

tyrro (approved):

olivierobert (approved):

sanG-github (approved):

PR #432, opened 2023-07-05
[#223] Ignore Danger run on WIP pull request

### LINKED ISSUE #223: Ignore Danger run on WIP pull request
We should ignore the Danger run on WIP pull requests to avoid some noise to the developer at the time that the PR is not ready for review.

Example:

```ruby
# Make it more obvious that a PR is a work in progress and shouldn't be merged yet.
has_wip_label = github.pr_labels.any? { |label| label.include? "WIP" }
has_wip_title = github.pr_title.include? "[WIP]"

if has_wip_label || has_wip_title
  message("Skipping Danger since PR is classed as Work in Progress")
  return
end
```

The code should handle: [WIP], WIP, label WIP, PR status `Draft`

mosharaf13:
close #223

## What happened 👀

- Skip danger run if PR has `[WIP]` in it's title
- Skip danger run if PR has `wip` in it's label
- Skip danger run if PR is in `draft`

## Insight 📝

- Tested on a project created from rails template. Worked successfully. 

## Proof Of Work 📹

Screen Shot 2023-07-05 at 3 08 43 PM

### DIFF
--- a/.template/addons/github/.github/workflows/test.yml.tt
+++ b/.template/addons/github/.github/workflows/test.yml.tt
@@ -2,7 +2,6 @@ name: Test
 
 on:
   pull_request:
-    types: [opened, edited, labeled, unlabeled, converted_to_draft, ready_for_review]
   push:
     branches:
       - develop
--- a/.template/addons/github/.github/workflows/test.yml.tt
+++ b/.template/addons/github/.github/workflows/test.yml.tt
@@ -2,6 +2,7 @@ name: Test
 
 on:
   pull_request:
+    types: [opened, edited, labeled, unlabeled, converted_to_draft, ready_for_review]
   push:
     branches:
       - develop
--- a/.template/addons/github/.github/workflows/test.yml.tt
+++ b/.template/addons/github/.github/workflows/test.yml.tt
@@ -144,7 +144,7 @@ jobs:
     needs: unit_tests
     runs-on: ubuntu-latest
     timeout-minutes: 5
-    if: github.event_name == 'pull_request'
+    if: github.event_name == 'pull_request' && !contains(join(github.event.pull_request.labels.*.name, ','), 'wip') && !contains(github.event.pull_request.title, '[wip]') && !github.event.pull_request.draft
     steps:
       - uses: actions/checkout@v3
         with:
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,13 +1,12 @@
 # frozen_string_literal: true
 
 # Skip Danger run if the opened pull request has "wip" label, [WIP] in the title, or status is Draft
-has_wip_label = github.pr_labels.any? { |label| label.include? 'wip' }
-has_wip_title = github.pr_title.include? '[WIP]'
+has_wip_label = github.pr_labels.any? { |label| label.downcase.include? 'wip' }
+has_wip_title = github.pr_title.downcase.include? '[wip]'
 is_draft = github.pr_draft?
 
 if has_wip_label || has_wip_title || is_draft
-  message('Skipping Danger since PR is classed as Work in Progress')
-  return
+  return message('Skipping Danger since PR is classed as Work in Progress')
 end
 
 # Runs Rubocop and submit comments on modified and added files
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,5 +1,11 @@
 # frozen_string_literal: true
 
+# Skip Danger run if the opened pull request has "wip" label, [WIP] in the title, or status is Draft
+if github.pr_labels.include?('wip') || github.pr_title.include?('[WIP]') || github.pr_json['draft']
+  message('Skipping Danger run: Pull request is marked as work-in-progress or draft.')
+  return
+end
+
 # Runs Rubocop and submit comments on modified and added files
 rubocop.lint(inline_comment: true, force_exclusion: true)
 
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,8 +1,12 @@
 # frozen_string_literal: true
 
 # Skip Danger run if the opened pull request has "wip" label, [WIP] in the title, or status is Draft
-if github.pr_labels.include?('wip') || github.pr_title.include?('[WIP]') || github.pr_json['draft']
-  message('Skipping Danger run: Pull request is marked as work-in-progress or draft.')
+has_wip_label = github.pr_labels.any? { |label| label.include? 'wip' }
+has_wip_title = github.pr_title.include? '[WIP]'
+is_draft = github.pr_draft?
+
+if has_wip_label || has_wip_title || is_draft
+  message('Skipping Danger since PR is classed as Work in Progress')
   return
 end
 
--- a/Dangerfile
+++ b/Dangerfile
@@ -6,7 +6,8 @@ has_wip_title = github.pr_title.downcase.include? '[wip]'
 is_draft = github.pr_draft?
 
 if has_wip_label || has_wip_title || is_draft
-  return message('Skipping Danger since PR is classed as Work in Progress')
+  message('Skipping Danger since PR is classed as Work in Progress')
+  return
 end
 
 # Runs Rubocop and submit comments on modified and added files
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,15 +1,5 @@
 # frozen_string_literal: true
 
-# Skip Danger run if the opened pull request has "wip" label, [WIP] in the title, or status is Draft
-has_wip_label = github.pr_labels.any? { |label| label.downcase.include? 'wip' }
-has_wip_title = github.pr_title.downcase.include? '[wip]'
-is_draft = github.pr_draft?
-
-if has_wip_label || has_wip_title || is_draft
-  message('Skipping Danger since PR is classed as Work in Progress')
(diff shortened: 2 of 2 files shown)


Dangerfile
```diff
@@ -1,5 +1,11 @@
 # frozen_string_literal: true
 
+# Skip Danger run if the opened pull request has "wip" label, [WIP] in the title, or status is Draft
+if github.pr_labels.include?('wip') || github.pr_title.include?('[WIP]') || github.pr_json['draft']
```
sanG-github, line 4: `github.pr_labels.include?('wip')`

The label could be: `type : wip` -> it does not match with this check. I think using Thanh's suggestion is safety.

```ruby
github.pr_labels.any? { |label| label.include? "wip" }
```

sanG-github: Also, we can call the `github.pr_draft?` directly 😜 [(ref)](https://github.com/danger/danger/blob/master/lib/danger/danger_core/plugins/dangerfile_github_plugin.rb#L151C9-L153)

mosharaf13: Yeah. That makes more sense. Fixed in 9a74913d2b9b861f7abb0fcb1d0635036a3937a1

Dangerfile
```diff
@@ -1,5 +1,15 @@
 # frozen_string_literal: true
 
+# Skip Danger run if the opened pull request has "wip" label, [WIP] in the title, or status is Draft
+has_wip_label = github.pr_labels.any? { |label| label.include? 'wip' }
+has_wip_title = github.pr_title.include? '[WIP]'
```
malparty, line 5:
```suggestion
has_wip_label = github.pr_labels.any? { |label| label.downcase.include? 'wip' }
has_wip_title = github.pr_title.downcase.include? '[wip]'
```

Should we make it case-insensitive? :)

Dangerfile
```diff
@@ -1,5 +1,15 @@
...
+is_draft = github.pr_draft?
+
+if has_wip_label || has_wip_title || is_draft
+  message('Skipping Danger since PR is classed as Work in Progress')
+  return
```
malparty, line 10:
```suggestion
  return message('Skipping Danger since PR is classed as Work in Progress')
```

Does it work in a single line? 👀

Dangerfile
mosharaf13: Thanks 🙏🏼 . Fixed in 7e57cc5674f5e5e9916ce4094835242822a179f1

Dangerfile
mosharaf13: Thanks 🙏🏼 . Fixed in 7e57cc5674f5e5e9916ce4094835242822a179f1

mosharaf13: Hi, Doing this rubocop throws error `Lint/TopLevelReturnWithArgument: Top level return with argument detected.`. Reverting back to previous implementation.

sanG-github (approved):

Dangerfile
```diff
@@ -1,5 +1,15 @@
 # frozen_string_literal: true
 
+# Skip Danger run if the opened pull request has "wip" label, [WIP] in the title, or status is Draft
+has_wip_label = github.pr_labels.any? { |label| label.downcase.include? 'wip' }
```
tyrro, line 4: @mosharaf13 , I was wondering whether we should move these conditions to the GH-action: there is already an `if` condition existing there:

https://github.com/nimblehq/rails-templates/blob/46573c27d4cbce5dcb9c1f8d3ff2984256f44799/.template/addons/github/.github/workflows/test.yml.tt#L147

I  am wondering, instead of mixing the logic between `test.yml` and `Dangerfile`, whether it's a good idea to keep it in a single place -  either in `test.yml` or in `Dangerfile`. As `test.yml` already has an `if` condition, we can add our new conditionals there. The updated `if` condition might look like:

```yml
if: github.event_name == 'pull_request' && contains( github.event.pull_request.labels.*.name, 'wip')  && contains( github.event.pull_request.title.*.name, 'wip') && github.event.pull_request.draft == false
```

@malparty ,  @sanG-github ,  would love to hear what you guys think of it as well, as you already commented on this PR!

Dangerfile
malparty: Oh, thanks for the sharing 🙏

Dangerfile
malparty: Moving the condition into the Github action has the advantage to not add any "GitHub" specific code in the DangerFile (which reduce cleaning required when a team is using GitLab or BitBucket) 🤔 

So I agree with this suggestion :)

🙈  The condition should be reversed though:
```
if: github.event_name == 'pull_request' && !contains( github.event.pull_request.labels.*.name, 'wip')  && !contains( github.event.pull_request.title.*.name, 'wip') && github.event.pull_request.draft == false
```

mosharaf13: Thanks for the suggestion 🙏🏼 . Fixed in d57cfc1d3a4a77beb4b093983ad64fa41ddcc0e3

malparty (approved):

Dangerfile
sanG-github: What a nice suggestion!! Thank Rajib 😘

.template/addons/github/.github/workflows/test.yml.tt
```diff
@@ -144,7 +144,7 @@ jobs:
     needs: unit_tests
     runs-on: ubuntu-latest
     timeout-minutes: 5
-    if: github.event_name == 'pull_request'
+    if: github.event_name == 'pull_request' && !contains(join(github.event.pull_request.labels.*.name, ','), 'wip') && !contains(github.event.pull_request.title, '[wip]') && !github.event.pull_request.draft
```
tyrro, line 147: Does it work without `join`ing? According to the Doc, this should work, no?
```suggestion
    if: github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name), 'wip') && !contains(github.event.pull_request.title, '[wip]') && !github.event.pull_request.draft
```

> contains(github.event.issue.labels.*.name, 'bug') returns true if the issue related to the event has a label "bug".

Reference: https://docs.github.com/en/actions/learn-github-actions/expressions#example-using-an-object-filter

tyrro, line 147: Also, could you check the case sensitivity here? For example, if we add `WIP` label in the PR instead of `wip`,  will it work?

tyrro, line 147: I have one more question - could you clarify that if we add the label `WIP` **after** creating the PR, does this workflow trigger successfully?

mosharaf13: I was trying to handle labels with wip substring like `type: wip`. You can see [here](https://github.com/mosharaf13/fleet-searcher-ruby/actions/runs/5530101137) skipping join tries to run danger while [here](https://github.com/mosharaf13/fleet-searcher-ruby/actions/runs/5530211236) it skipped danger for same label

mosharaf13: Yeah. As you can see [here](https://github.com/mosharaf13/fleet-searcher-ruby/pull/16), this check not case sensitive for labels. 

Refererence: https://docs.github.com/en/actions/learn-github-actions/expressions#contains

mosharaf13: Umm, no. This check triggers only when creating the PR. 

To trigger the workflow each time a label or title has been changed or PR status changed to/from draft we will have to add change shown below.

Should I add these changes? 


```
on:
  pull_request:
    types: [opened, edited, labeled, unlabeled, converted_to_draft, ready_for_review]
  push:
    branches:
      - develop
      - main
```

sanG-github: IMO, I think it's good idea to add it @mosharaf13. 😉

mosharaf13: Thanks 👍🏼 . Added in b925ebc529f1ade2688f5ad2e730e94ced7a42b9

tyrro: ah got it, `type: wip` is a good example why we need `join` here!

tyrro: Thanks for the confirmation! Good to know that they are not case-sensitive!

tyrro (approved):

.template/addons/github/.github/workflows/test.yml.tt
```diff
@@ -2,6 +2,7 @@ name: Test
 
 on:
   pull_request:
+    types: [opened, edited, labeled, unlabeled, converted_to_draft, ready_for_review]
```
olivierobert, line 5: Why do we need to be so specific? Can't this be triggered just with `pull_request`?

mosharaf13: Thanks for catching the issue. Come to think of it, we don't really need an immediate workflow trigger after a PR state change, as no code has been changed.

And, just `pull_request` gets the job done where we make an existing open PR WIP and push commits in it; for each subsequent commit, it skips danger.

Fixed in d5fc88d7a3d991bf619fda410262669c31c03a38

malparty (approved):

PR #443, opened 2023-07-11
Remove the ./bin/dev.sh script in favor of the makefile

### LINKED ISSUE #410: Remove the `./bin/dev.sh` script in favor of the makefile
## Why

Currently, the `make dev` command is more complex than it should be:
`make dev` calling `./bin/dev.sh` calling `foreman start -f Procfile.dev` calling `rails s` 🥵 

Let's remove `./bin/dev.sh` and bring `foreman start -f Procfile.dev` directly in the make file 🚀 

## Who Benefits?

Dev when looking for what commands must run behind the hood.

mosharaf13:
close #410

## What happened 👀

- Move `foreman start -f Procfile.dev` from `./bin/dev.sh` to Makefile
- Remove `./bin/dev.sh`

## Insight 📝

-

## Proof Of Work 📹

The server starts without any error
Screen Shot 2023-07-11 at 1 57 27 PM


malparty (approved):
🚀

tyrro (approved):
As I am looking at the `template_spec.rb` in this PR, I came across another spec file that's named as `template.rb`, which should be `template_spec.rb` as per Rails convention:

https://github.com/nimblehq/rails-templates/blob/704ae5e347cf012eb63aa041ff4fc0a158fa446b/.template/spec/addons/base/phrase/template.rb

could you rename the above file to `template_spec.rb`, this is not related to this PR; but this is a very small change and makes the naming of test file consistent!

sanG-github (approved):

malparty (approved):

PR #452, opened 2023-08-01
[#413] The test coverage does NOT show the correct statistics

### LINKED ISSUE #413: Fix: The test coverage does NOT show the correct statistics
## Issue

The code coverage always shows 0.00%, which means it does not reflect the real code coverage of the project. It was caused by the SimpleCov not starting before the Rails was loaded. 
- Sample Fix on [EwaPayroll](https://diffy.org/diff/114f610cff391)
- [Related issue](https://github.com/simplecov-ruby/simplecov/issues/77)

image

## Expected

Show accurate statistics.
image

## Steps to reproduce

1. Generate the project from the template
2. Push the code to the remote repository
3. Add codes, and create a PR
4. 💥

sanG-github:
close #413 

## What happened 👀

Resolved the issue related to the code coverage always showing 0.00%, which means it does not reflect the real code coverage of the project. 

## Insight 📝

It was caused by the SimpleCov not starting before the Rails was loaded.

## Proof Of Work 📹

Tested and worked well on [the public repo](https://github.com/sanG-github/test-code-coverage/pull/1#issuecomment-1660059827)

Screenshot 2023-08-01 at 17 57 12

### DIFF
--- a/.template/addons/github/.github/workflows/test.yml.tt
+++ b/.template/addons/github/.github/workflows/test.yml.tt
@@ -74,7 +74,7 @@ jobs:
         run: docker-compose pull test || true
 
       - name: Run codebase test
-        run: docker-compose run test bundle exec rspec spec/codebase/codebase_spec.rb --format progress
+        run: docker-compose run test -e DISABLE_SPRING=1 bundle exec rspec spec/codebase/codebase_spec.rb --format progress
 
   unit_tests:
     name: Unit tests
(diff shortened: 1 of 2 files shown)


tyrro (approved):

malparty (approved):

.template/addons/github/.github/workflows/test.yml.tt
```diff
@@ -74,7 +74,7 @@ jobs:
         run: docker-compose pull test || true
 
       - name: Run codebase test
-        run: docker-compose run test bundle exec rspec spec/codebase/codebase_spec.rb --format progress
+        run: docker-compose run test -e DISABLE_SPRING=1 bundle exec rspec spec/codebase/codebase_spec.rb --format progress
```
andyduong1920, line 77: Hi @sanG-github we are facing an error due to this changes in our new project, could you double check that, we removed the `-e DISABLE_SPRING=1` now in our new project, thank you

image

PR #457, opened 2023-08-04
[#456] Fix setting Chrome's locale to non-English will cause some Capybara tests to fail

### LINKED ISSUE #456: Fix: Setting Chrome's locale to non-English will cause some Capybara tests to fail
## Issue

Capybara tries to run the UI tests on different locale than our test suite causing a problem looking up some elements which contain localized string.

## Expected

Capybara should not inherit the page locale from the local web browser locale and instead always use English.

## Steps to reproduce

1. Set Chrome's locale to non-English language such as Thai.
2. Write a test which select an option from a select box field such as `select I18n.t('activerecord.attributes.model.field.example_option'), from: 'model[field]'`
3. Run the test `CHROME_HEADLESS=false bundle exec rspec FILENAME`.
4. It will raise this error.
```
Failure/Error: select I18n.t('activerecord.attributes.model.field.example_option'), from: 'model[field]'
     
Capybara::ElementNotFound:
     Unable to find option "Example option" within #<Capybara::Node::Element tag="select" path="/HTML/BODY[1]/MAIN[1]/DIV[1]/FORM[1]/DIV[1]/SELECT[1]">
```

topnimble:
close #456 

## What happened 👀

Capybara tries to run the UI tests on different locale than our test suite causing a problem looking up some elements which contain localized string.

## Insight 📝

- Capybara should not inherit the page locale from the local web browser locale and instead always use English.

## Proof Of Work 📹

All tests passed.

### DIFF
--- a/.template/variants/web/spec/support/capybara.rb
+++ b/.template/variants/web/spec/support/capybara.rb
@@ -28,6 +28,9 @@ Capybara.register_driver(:headless_chrome) do |app|
   # Disable /dev/shm use in CI
   options.add_argument('disable-dev-shm-usage') if ENV['CI']
 
+  # Use English locale for displaying page
+  options.add_preference('intl.accept_languages', 'en')
+
   Capybara::Selenium::Driver.new(
     app,
     browser: :chrome,


sanG-github (approved):

andyduong1920:
@topnimble I think you need to assign this PR to yourself, not to Xavier 🤞

andyduong1920 (approved):

malparty (approved):
🚀

malparty:
> > @topnimble I think you need to assign this PR to yourself, not to Xavier 🤞
> 
> Xavier was the person who actually wrote this patch so it would be better to assign his name instead of me 👍.

You owned it: Not only you were the one to face the issue on your machine, but you took time to write the issue and prepare this PR 💪 
Thanks for the intention though, I appreciate it 😇

PR #458, opened 2023-08-07
Update the sidekiq queue config

andyduong1920:
## What happened 👀

Add the queues with Rails env prefix in order to allow the Sidekiq job get processed.

## Insight 📝

The new default Rails generate had the `config.active_job.queue_name_prefix = Rails.env` which will require the Sidekiq queue to include these env prefixes.

## Proof Of Work 📹

Used in client projects

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -3,3 +3,7 @@
 :queues:
   - [default, 1]
   - [mailers, 2]
+  - [development_default, 1]
+  - [development_mailers, 2]
+  - [production_default, 1]
+  - [production_mailers, 2]
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -1,8 +1,6 @@
 ---
 :concurrency: <%= ENV['SIDEKIQ_CONCURRENCY'] || 4 %>
 :queues:
-  - [default, 1]
-  - [mailers, 2]
   - [development_default, 1]
   - [development_mailers, 2]
   - [production_default, 1]


sanG-github (commented):
I'm just curious if there are any benefits from this prefix since we seem never to run our application with both environments on the same machine. 🤔 

```ruby
config.active_job.queue_name_prefix = nil
```

andyduong1920:
> I'm just curious if there are any benefits from this prefix since we seem never to run our application with both environments on the same machine. 🤔
> 
> ```ruby
> config.active_job.queue_name_prefix = nil
> ```

The only benefit I can see is it allow us to have different setup per env like

- development_default: 1
- production_default: 3

olivierobert:
@sanG-github comment makes sense as the use case it covers does not exists for us as all environments are isolated. @Goose97 brought this into his IC but now it does not make sense to upstream this IMO 😅

andyduong1920:
@sanG-github @olivierobert But by default, the Rails generator will set the `queue_name_prefix` as Rails.env in the `config/application.rb`.

```
    # Prefix the queue name of all jobs with Rails ENV
    config.active_job.queue_name_prefix = Rails.env
```

So if we do not modify the `queue` here, the new application generated from our template won't work, and it requires the user to manually update the `config/sidekiq.yml`.

Incase we do not need this PR, we would need to open another PR in our template to - update the `config.active_job.queue_name_prefix` in `config/application.rb` to `nil`

```
    # Prefix the queue name of all jobs with Rails ENV
    config.active_job.queue_name_prefix = nil
```

cc @malparty @Goose97

malparty:
@olivierobert @sanG-github to clarify:

These new queue names (with env prefix) are the default generated by the `rails new` command. 

So there are 2 options:
- Keep the names with the env prefix, with no deviation from a standard Rails project (this PR suggestion)
- Edit our template so that it will edit the `config/application.rb` file generated by `rails new` command to replace these lines:

```
    # Prefix the queue name of all jobs with Rails ENV
    config.active_job.queue_name_prefix = Rails.env
```

In my opinion, it's better to reduce the deviation with Rails defaults (less code, less maintenance to handle). So I recommend moving forward with this PR 👍

config/sidekiq.yml
```diff
@@ -3,3 +3,7 @@
 :queues:
   - [default, 1]
   - [mailers, 2]
```
malparty, line 5: Should we remove these 2 queues then? 🤔

andyduong1920: Yeah we can remove it as it's useless now 👍

andyduong1920: removed in https://github.com/nimblehq/rails-templates/pull/458/commits/a8868a7757acc526346bae0d11e857a8449d02cf

malparty (approved):

sanG-github (approved):

sanG-github:
@andyduong1920 

Please help to resolve the failing CI, then we're good to go!!! 🚀

malparty:
@andyduong1920 I just rebased that PR onto develop, that should fix the failing test 😇

andyduong1920:
> @andyduong1920 I just rebased that PR onto develop, that should fix the failing test 😇

Thank you 🚀

olivierobert (approved):

Issue #446, opened 2023-07-25
Add OpenAPI to support creating API documentation with OpenAPI


suho:
## Why

The team now has [OpenAPI Training Path](https://notion.so/d76224c3480242aeb4e8c71055a76398?pvs=25) to learn and practice. [OpenAPI on Compass](https://github.com/nimblehq/compass/pull/1283) for developers to follow conventions and best practices. NO OpenAPI boilerplates currently exist in the rails templates to support web developers in creating API documentation with OpenAPI.

## Acceptance Criteria

- Add OpenAPI addon as a required addon in the template.
- Add OpenAPI tooling (Spectral, Elements, Lint).
- Create an endpoint (e.g., `GET {{base_url}}/api/docs/openapi`) to access the API documentation through the browser.
- Add Mock Server with Prism (related to this [API Mock Server with OpenAPI](https://www.notion.so/API-Mock-Server-with-OpenAPI-83fda1653eb1409fab9b08c1fcf466fd?pvs=21)).
- Add CI/CD workflow to run linting and deploy mock server to fly.io.

## Resources

- [Add OpenAPI to Web templates](https://www.notion.so/nimblehq/Add-OpenAPI-to-Web-templates-5b843ba050464edb80d14dfbad6da0fe?pvs=4)
- https://github.com/nimblehq/api-mock-templates/pull/3
- https://github.com/nimblehq/nimble-survey-web

## Who Benefits?

Everyone

Issue #422, opened 2023-07-03
Change to use `@nimblehq/eslint-config-nimble-core`


andyduong1920:
## Why

- As the `@nimblehq/eslint-config-nimble` team will move forward with the `@nimblehq/eslint-config-nimble-core`
- But there is an issue with the @nimblehq/eslint-config-nimble-core at the writing time, https://github.com/nimblehq/eslint-config-nimble/issues/56 would need the team solve it first

## Who Benefits?

Developers

PR #464, opened 2023-08-31
[#446][Part 1] Add OpenAPI to support creating API documentation with OpenAPI

sanG-github:
Implement #446

## What happened 👀

- Added OpenAPI add-on.
- Added an example for OpenAPI spec (/health API) - in the folder docs/openapi.
- Added tools for OpenAPI (e.g., spectral, eslint)
- Added tests to cover cases for OpenAPI add-on.
- Added lint - CI workflow for running lint on OpenAPI spec.

## Insight 📝

This is the first part of the implementation, I intend to implement:
- Part 2: Add an endpoint (/api/docs/openapi) to support generating OpenAPI docs by using Spotlight
- Part 3: Add mock server

## Proof Of Work 📹

Generated the project successfully
image

The generated project
image


.gitignore.rb
```diff
@@ -10,5 +10,8 @@
 
     # Ignore the test coverage results from SimpleCov
     /coverage
+
+    # Ignore generated OpenAPI file
+    /public/openapi.yml
```
malparty, line 15: This should only be added if the user selected the `Open API` addon. Let's create a `.gitignore.rb` file inside the `addons/openapi` folder 👍 

(example of implementation is available as a similar implementation is done for the web variant: `.template/variants/web/.gitignore.rb`)

.template/addons/github/.github/workflows/lint_docs.yml.tt
malparty, line 1: This would be added when the user selects the GitHub addons. 

But I think we don't want it when we do NOT select the OpenAPI addon. 

In this case, should we have this file inside the `OpenAPI` addon folder, and use a ruby command to copy the file ONLY if there is a `.github/workflows` folder already created? (we also need to ensure the  OpenAPI addons will be setup AFTER the GitHub addons).

.spectral.yml
malparty, line 1: Could you explain why this file is at the root of the template? As far as I know, the template does not need any JSON parsing, does it? 👀

.template/addons/openapi/docs/openapi/examples/requests/.keepfile
malparty, line 1: Let's rename it `.keep` for consistency with the other `.keep` files.

malparty: (same for all the other `.keepfile` files in this PR)

.template/variants/api/package.json.rb
```diff
@@ -0,0 +1,27 @@
...
+run 'yarn add --dev @stoplight/spectral-cli@6.8.0'
+run 'yarn add --dev eslint-plugin-yml@1.8.0'
+
+# Setup scripts
+run 'npm pkg set scripts.lint:docs:yml="eslint docs/openapi --ext .yml --color"'
+run 'npm pkg set scripts.lint:docs:openapi="spectral lint docs/openapi/openapi.yml -F error"'
+run 'npm pkg set scripts.lint:docs:dev="yarn lint:docs:yml && yarn lint:docs:openapi"'
+run 'npm pkg set scripts.lint:docs:public="yarn build:docs && eslint public/openapi.yml --color --no-ignore && spectral\
+ lint public/openapi.yml -F error"'
+run 'npm pkg set scripts.build:docs="swagger-cli bundle docs/openapi/openapi.yml --outfile public/openapi.yml --type\
```
malparty, line 26: If "OpenAPI" is an addons, this should not be inside the API variant folder, but only inside the openapi addons folder, no?

.template/variants/web/package.json.rb
```diff
@@ -40,6 +43,13 @@
 run 'npm pkg set scripts.stylelint:fix="stylelint **/*.scss --color --fix"'
 run 'npm pkg set scripts.codebase="yarn eslint && yarn stylelint"'
 run 'npm pkg set scripts.codebase:fix="yarn eslint:fix && yarn stylelint:fix"'
+run 'npm pkg set scripts.lint:docs:yml="eslint docs/openapi --ext .yml --color"'
+run 'npm pkg set scripts.lint:docs:openapi="spectral lint docs/openapi/openapi.yml -F error"'
+run 'npm pkg set scripts.lint:docs:dev="npm lint:docs:yml && npm lint:docs:openapi"'
+run 'npm pkg set scripts.lint:docs:public="npm build:docs && eslint public/openapi.yml --color --no-ignore && spectral\
+ lint public/openapi.yml -F error"'
+run 'npm pkg set scripts.build:docs="swagger-cli bundle docs/openapi/openapi.yml --outfile public/openapi.yml --type\
+ yaml"'
```
malparty, line 52: Same here, also having these command added from the `addons/openapi` will allow to dry this code.

malparty (changes requested):
Let's be sure to not "pollute" the projects that will not select the OpenAPI addons in the future. We should only add content from the openapi addons scripts. Not from the Web or API variant 🙏

.spectral.yml
hoangmirs: Yes, this should be included when we selecting the API variant only 🤔

.template/addons/github/.github/workflows/lint_docs.yml.tt
hoangmirs: So we should make sure that this addon will be run after `github` addon, right?

malparty: Yes, that's a good point.

sanG-github: @suho 
We already have a discussion about this issue, right? Since we need to add `OpenAPI` as a mandatory addon instead of an optional one. Could you please help to clarify and elaborate on that to help reduce the confusion? Thank you in advance.

malparty: Will clarified when Su is back. PR: if mandatory, it's not an add-on anymore ^^ it would be part of the core (even before the call for API/Web variant)

suho: OpenAPI is mandatory for the API variant. (I mentioned in [the AC](https://github.com/nimblehq/rails-templates/issues/446)), also, Olivier [mentioned about this](https://github.com/nimblehq/gin-templates/pull/106#discussion_r1301111294) when I was working on the gin-templates

malparty: okay, so that mean it is still an optional one for the Web variant and so it's worth keeping it isolated with the rest of the code (as any other `add-on`) :)

Makefile
```diff
@@ -1,9 +1,10 @@
+# Y - in response to Would you like to add the OpenAPI addon?
```
suho, line 1: @sanG-github I think we don't need this options, just automatically add `OpenAPI` whenever the variant is API

sanG-github: This `Makefile` is for test purposes only, it doesn't impact our template flow. @suho

suho: @sanG-github While it doesn't impact our template, it's unnecessary to have this change as we did not ask the question "Would you like to add the OpenAPI addon?"

sanG-github: Ah okay I got the idea, updated it in 71c5dd2 ✅

suho (approved):

.template/addons/github/.github/workflows/lint_docs.yml.tt
malparty, line 1: Let's move this file into `.template/addons/openapi/.github/workflows/lint_docs.yml`

From my understanding, this file does not need to be a `.tt` as it is not using any template formating instructions. 

Then, we can have an instruction in `.template/addons/openapi/template.rb` to copy the file `if install_github` (variable to add).


I recommend this so the GitHub add-on does not become too complex over time. My main concern is that the GitHub add-on ends up with too many add-files, remove files, etc... 

This also mean we might want to run the GitHub add-on before the OpenAPI add-on (which would be mandatory if later we want to add some wiki instructions/content specific to OpenAPI 👍 )

.template/addons/openapi/package.json.rb
```diff
@@ -0,0 +1,27 @@
...
+                  "private": "true"
+                }
+              JSON
+end
+
+run 'yarn add --dev eslint'
+run 'yarn add --dev @nimblehq/eslint-config-nimble@2.2.1'
+run 'yarn add --dev @apidevtools/swagger-cli@4.0.4'
+run 'yarn add --dev @stoplight/spectral-cli@6.8.0'
+run 'yarn add --dev eslint-plugin-yml@1.8.0'
```
malparty, line 18:
```suggestion
run 'yarn add --dev @nimblehq/eslint-config-nimble@latest'
run 'yarn add --dev @apidevtools/swagger-cli@4.0'
run 'yarn add --dev @stoplight/spectral-cli@6.8'
run 'yarn add --dev eslint-plugin-yml@1.8'
```

Should we:
- use `latest` for our ESLint config? (if yes, please let's also update the Web variant 🙇 )
- use the latest patch version for the external packages? 🤔

malparty (changes requested):

.template/addons/github/.github/workflows/lint_docs.yml.tt
sanG-github: That's really nice suggestion @malparty. 

I tried locally and then realized that if we implement by the way you suggested, it still requires `GitHub` addon to be added before the `OpenAPI`. 

Thus, we need to re-order the flow of setting up the addons, so I think that would be better to implement it in #473  as we need to introduce `@install_github`, please let me know your thoughts on it.

.template/addons/openapi/package.json.rb
sanG-github: Thanks for pointing this out, I updated it in a943e33 ✅

.template/addons/github/.github/workflows/lint_docs.yml.tt
malparty: I would argue that #473 will have too many changes, so if we can avoid adding some additional work with this PR, it's better.

And also because if we merge it this way, it might give the wrong example for other contributions, in how we distribute the responsibilities 🤔

sanG-github: Thanks for your information, I updated it in 27edf91 ✅

.template/addons/openapi/package.json.rb
malparty: Okay, let's keep 2.2.1. I will create an issue to have a dedicated PR: the current version is 2.6, there are probably many lint to be fixed manually 🙈

malparty: Actually there are already stories related with that: https://github.com/nimblehq/rails-templates/issues/422

Makefile
```diff
@@ -1,3 +1,6 @@
...
+openapi_addon_prompt = Y\n
+
 # Y - in response to Would you like to add the Github addon?
```
malparty, line 4: I think it's better if we stop using there variables (common_addonx_prompts, web/api, ...) and instead just use a string and have the comments before:

```
# Y - in response to Would you like to add the Github addon?
# Y - in response to Would you like to add the Open[API](https://docs.github.com/) addon?
# Y - in response to Would you like to add the SemaphoreCI addon?
# Y - in response to Would you like to add the Nginx addon?
# Y - in response to Would you like to add the Phrase addon?
# Y - in response to Would you like to add the Devise addon?
# Y - in response to Would you like to add the Bootstrap addon?
# Y - in response to Would you like to add the Slim Template Engine addon?
# Y - in response to Would you like to add the Hotwire addon?
create_web:
	printf "Y\nY\nY\nY\nY\nY\nY\nY\nY\n" | rails new $(APP_NAME) -m ./template.rb -T ${OPTIONS}
```

(and duplicate that for the API)

Because how the API includes the OpenAPI addons, using core makes it hard to read. Also let's ensure to keep the new ordering (GitHub first)

Thanks :)

(that should be my last comment :bow:)

sanG-github: Just resolved your last comment in 5695671 ✅ 😛

malparty (approved):

olivierobert (approved):

PR #468, opened 2023-09-14
[#459] Remove schema.rb from .gitattributes linguist-generated

### LINKED ISSUE #459: Remove schema.rb from .gitattributes linguist-generated
## Why

There is not a single time that my teammates and I don't review this file in the PR.
In deed, 
- if a migration file was updated but the schema was not updated, this can result in issues.
- if the PR assignee ran the app on another PR (with migrations) and came back to its PR, this can include unwanted changes in the `schema.rb` file, again, prone to deployment errors or bugs

image

## Who Benefits?

Code reviewers who will have 1 click less to perform.

malparty:
close #459

## What happened 👀

- Remove db/schema.rb from the git attributes auto-hidden files
- Add a test as well

## Insight 📝

`n/a`

## Proof Of Work 📹

After generating a new app named `man` (for `manger` in French, meaning `to eat` 🤡):

image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# See https://git-scm.com/docs/gitattributes for more about git attribute files.
+
+# Mark any vendored files as having been vendored.
+vendor/* linguist-vendored
--- a/.template/spec/base/template_spec.rb
+++ b/.template/spec/base/template_spec.rb
@@ -30,4 +30,9 @@ describe 'Base template' do
   it 'creates Reek configuration files' do
     expect(file('.reek.yml')).to exist
   end
+
+  it 'creates the .gitattributes without schema.db' do
+    expect(file('.gitattributes')).to exist
+    expect(file('.gitattributes')).not_to contain('db/schema.rb')
+  end
 end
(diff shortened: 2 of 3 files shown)


olivierobert (approved):
This was a TIL for me :-) 

As an aside, Bitbucket does not support `gitattributes` yet (https://jira.atlassian.com/browse/BSERV-2912) while Gitlab seems to have some support for it (https://docs.gitlab.com/ee/user/project/git_attributes.html). So for now it is very GitHub specific.

.gitattributes
```diff
@@ -0,0 +1,4 @@
+# See https://git-scm.com/docs/gitattributes for more about git attribute files.
+
+# Mark any vendored files as having been vendored.
```
olivierobert, line 3: Did you mean:

```suggestion
# Mark any vendor files as having been vendored.
```

malparty: I just reused the official Rails file 🙈 

https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/gitattributes.tt

malparty: Said differently, I just removed the line with the schema file, to reduce any deviation with a fresh `rails new` project 👍

tyrro (approved):

sanG-github (approved):

PR #479, opened 2023-11-09
[#473] Revise the Flow for Asking Options to generate the project 

### LINKED ISSUE #473: Revise the Flow for Asking Options to Generate the Project
## Why

Currently, when generating a project, our workflow involves prompting the user to select add-ons for installation one at a time. After each selection, we immediately proceed to install the chosen add-on before moving on to the next question. This approach, while functional, has its drawbacks. It consumes more time than necessary and can result in a less-than-ideal user experience.

```
Would you like to add the Github addon? (y/N)
-> y
-> Install the Github addon

Would you like to add the SemaphoreCI addon? (y/N)
-> N

Would you like to add the Nginx addon?
-> y
-> Install the Nginx addon
```

To enhance this process, we should consider revising the flow so that we first present all the questions to the user before initiating any installations. This approach eliminates the need for the user to wait for individual installations before proceeding to the next question.

```
Would you like to add the Github addon? (y/N)
-> y
Would you like to add the SemaphoreCI addon? (y/N)
-> N
Would you like to add the Nginx addon?
-> y

-> Install the Github addon
-> Install the Nginx addon
```

By implementing this revised workflow, we can streamline the user experience and improve the efficiency of the add-on installation process.

## Who Benefits?

The user who generates the project from our template.

sanG-github:
close #473

## What happened 👀

Enhance the flow to generate the project by using `@instance_variable` to remember the user's options for each addon.
Then we install all the necessary addons based on these variables's values.

## Insight 📝

`n/a`

## Proof Of Work 📹

| Generating Web App                                                                                                                   | Generating API App                                                                                                                   |
|--------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------|
| image | image |


malparty (approved):
At first I wanted to suggest to move the `ask_for_optional_addons` call earlier, but then I saw it better for the UX to have the message about the default add-ons installations, so that it's clear what is already configure when the user select the optional addons. 

So all good! 

PR: just the proof of work is wrong for the Web variant, it displays the prompt of the API variant instead of:

image

olivierobert (approved):

longnd (approved):
much nicer :)

PR #481, opened 2023-11-10
[#478] Fix: Duplicated `Dockerfile` and `.dockerignore` when generating the project with Rails 7.1

### LINKED ISSUE #478: Fix: Duplicated `Dockerfile` and `.dockerignore` when generating the project with Rails 7.1
## Issue

We're facing two duplicating errors when generating the project from our template that should be fixed ASAP👷‍ 

image

## Expected

The template can generate a project successfully and there are no duplications or errors caused to the user

## Steps to reproduce

Both ways to generate the project faced the same issue.
```ruby
1. Run `rails new app_name -m https://raw.githubusercontent.com/nimblehq/rails-templates/main/template.rb` as documented.

Or

1. Clone our rails template locally
2. Run `rails new app_name -m ./template.rb`
```

malparty:
close #478

## What happened 👀

Just remove the files before applying the Docker template. Later we might want to reverse that if we choose to use the default Rails docker files 👍 

## Insight 📝

`n/a`

## Proof Of Work 📹

image

### DIFF
--- a/.template/addons/docker/template.rb
+++ b/.template/addons/docker/template.rb
@@ -2,6 +2,9 @@
 
 use_source_path __dir__
 
+remove_file 'Dockerfile'
+remove_file '.dockerignore'
+
 template 'Dockerfile.tt'
 template 'docker-compose.dev.yml.tt'
 template 'docker-compose.test.yml.tt'


sanG-github (approved):
Thanks for the quick fix and investigation 🚀

tednguyendev (approved):

PR #495, opened 2024-01-03
[#494] Create (or use) a Rubocop for the inverse keyword needed in the associations

### LINKED ISSUE #494: Create (or use) a Rubocop for the inverse keyword needed in the associations
## Why

We have a comprehensive set of style guides in the Compass. Sometimes, it hard to keep them all of your head why developing. Turning it into a custom Rubocop rules can lower the cognitive overhead

- Implement (or use) a custom Rubocop to catch the missing `inverse` keyword in Rails associations

## Who Benefits?

Developers on standard admin projects.

Goose97:
- Close #494

## What happened 👀

Implement custom cop to enforce explicit `inverse_of` when specifying Rails association
[Compass rule](https://nimblehq.co/compass/development/code-conventions/ruby/ruby-on-rails/#convention-13)

## Insight 📝

- [ ] Detect Rails model files and only run this cop on such files

There are severals trade-offs that I made to reduce the complexity of the cop:
- We can improve the performance by only scanning the top level expressions of the class. I haven't tested it yet, but I suspect that the gain would be negligible. The reason is that we use a lot of different cop rules so that there's a good chance that we have to scan the whole file anyway.
- There's no easy way to distinguish between Rails association helpers and our application functions. If the names collide, it might lead to false positive. However, in practice, we should avoid from naming our methods this way, so I think we could safely ignore this case.

```ruby
class Author < ActiveRecord::Model
  has_many :books # we should scan this one
  belongs_to :user # and this one

  def test
    SomeModule.has_many(...) # but not this one
  end
end
```

## Proof Of Work 📹

https://github.com/nimblehq/rails-templates/assets/35915460/eb7efad9-db43-4382-83e2-f63cdc08a778

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,6 +2,7 @@ require:
   - rubocop-rails
   - rubocop-rspec
   - rubocop-performance
+  - ./lib/custom_cops/required_inverse_of_relations.rb
 
 AllCops:
   Exclude:
@@ -50,3 +51,6 @@ RSpec/NestedGroups:
 
 RSpec/MultipleExpectations:
   Max: 10
+
+CustomCops/RequiredInverseOfRelations:
+  Enabled: true
--- a/.template/addons/custom_cops/template.rb
+++ b/.template/addons/custom_cops/template.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+require 'fileutils'
+
+FileUtils.mkdir_p 'lib/custom_cops'
+copy_file 'custom_cops/required_inverse_of_relations.rb', 'lib/custom_cops/required_inverse_of_relations.rb'
--- a/custom_cops/Gemfile
+++ b/custom_cops/Gemfile
@@ -1,0 +1,4 @@
+source 'https://rubygems.org'
+
+gem 'rubocop'
+gem 'rspec'
--- a/custom_cops/Gemfile.lock
+++ b/custom_cops/Gemfile.lock
@@ -1,0 +1,53 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+    ast (2.4.2)
+    diff-lcs (1.5.0)
+    json (2.7.1)
+    language_server-protocol (3.17.0.3)
+    parallel (1.24.0)
+    parser (3.2.2.4)
+      ast (~> 2.4.1)
+      racc
+    racc (1.7.3)
+    rainbow (3.1.1)
+    regexp_parser (2.8.3)
+    rexml (3.2.6)
+    rspec (3.12.0)
+      rspec-core (~> 3.12.0)
+      rspec-expectations (~> 3.12.0)
+      rspec-mocks (~> 3.12.0)
+    rspec-core (3.12.2)
+      rspec-support (~> 3.12.0)
+    rspec-expectations (3.12.3)
+      diff-lcs (>= 1.2.0, < 2.0)
+      rspec-support (~> 3.12.0)
+    rspec-mocks (3.12.6)
+      diff-lcs (>= 1.2.0, < 2.0)
+      rspec-support (~> 3.12.0)
+    rspec-support (3.12.1)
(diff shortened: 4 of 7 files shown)


Goose97:
@malparty Now think about it, I think we better off splitting the custom cops into a separate repo 🤔 . It's likely that we gonna have more custom cops coming in the future and also makes it easier for current project to adopt the new cops. What do you think?

.rubocop.yml
```diff
@@ -2,6 +2,7 @@ require:
   - rubocop-rails
   - rubocop-rspec
   - rubocop-performance
+  - ./custom_cops/required_inverse_of_relations.rb
```
malparty, line 5: I'm not sure if that works, but can we try to rename the folder: `./rubocop/cops/`. At first glance, `custom_cops` might be hard to recognize for people new in the world of Rails 🛤️

malparty (approved):
Thanks for the clear insights about the arbitrations made. That seems fair enough to me 💡 

Approved ahead with a question/suggestion 👍

.rubocop.yml
Goose97: Agree. Fixed in fba261b028c05b6a3fc2fa6e08450028ba607a89

rubocop/Gemfile
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+source 'https://rubygems.org'
+
+gem 'rspec'
+gem 'rubocop'
```
sanG-github, line 6: I'm curious about this one @Goose97, wasn't it added in `Gemfile.tt` inside the `group :development, :test` scope? 🤔

.rubocop.yml
```diff
@@ -2,6 +2,7 @@ require:
   - rubocop-rails
   - rubocop-rspec
   - rubocop-performance
+  - ./rubocop/cops/required_inverse_of_relations.rb
```
sanG-github, line 5: As we're using `CustomCops/RequiredInverseOfRelations` for the rule's name, should we use it instead? 🤔 

```suggestion
  - ./rubocop/custom_cops/required_inverse_of_relations.rb
```

rubocop/cops/required_inverse_of_relations.rb
```diff
@@ -0,0 +1,35 @@
...
+    def_node_matcher :association_expression_with_arguments?, <<~PATTERN
+      (send nil? ASSOCIATION_METHODS (sym _) (hash $...))
+    PATTERN
+
+    def on_send(node)
+      return add_offense(node) if association_expression_no_arguments?(node)
+
+      return unless (hash_pairs = association_expression_with_arguments?(node))
+
+      add_offense(node) unless contain_inverse_of?(hash_pairs)
```
sanG-github, line 24: It's more readable I guess, and using `if` reduces cognitive loads when reading the code 😂 

```suggestion
      hash_pairs = association_expression_with_arguments?(node)

      add_offense(node) if hash_pairs && contain_inverse_of?(hash_pairs)
```

sanG-github (approved):
What a delightful change!! 🎉

rubocop/Gemfile
Goose97: this Gemfile is for testing the custom rubocop, unlike the Gemfile.tt that will be created during the template bootstrap. In another word, the users of this template won't see this file, only us the developers 👍

.rubocop.yml
Goose97: Good point. It's not a strict requirement to follow the naming convention of Zeitwerk here as we manually requires the file. However, it makes it easier to locate the file. That's a win for sure

fd294bb0350c2dd9e4151390b5efd7533d704a34

rubocop/cops/required_inverse_of_relations.rb
Goose97: You're spot on. Fixed in fbdd5c688803fd1af1b92c3c8245029c07ae8c18

PR #497, opened 2024-01-03
[#461] Move user_email in publish_wiki workflow to a secret

### LINKED ISSUE #461: Move user_email in publish_wiki workflow to a secret
## Why

- Currently, we are putting the user_email into the workflow with plain text.
- As this is a part of credentials, we should create a secret for it:
https://github.com/nimblehq/rails-templates/blob/26835203353ef6fc0ccf2ee079c010d17c6ab8ff/.template/addons/github/.github/workflows/publish_wiki.yml.tt#L30
- `user_email` should be put in the secrets like the `user_access_token`

## Who Benefits?

Developers

sanG-github:
- Close #461

## What happened 👀

Move `user_email` in the `publish_wiki` workflow to a `secret` instead of publicizing it.

## Insight 📝

`n/a`

## Proof Of Work 📹

Just changed the usage, nothing special. 🚀

### DIFF
--- a/.template/addons/github/.github/workflows/publish_wiki.yml.tt
+++ b/.template/addons/github/.github/workflows/publish_wiki.yml.tt
@@ -27,5 +27,5 @@ jobs:
         uses: nimblehq/publish-github-wiki-action@v1.0
         with:
           user_name: github-wiki-workflow
-          user_email: bot@nimblehq.co
+          user_email: ${{ secrets.USER_EMAIL }}
           user_access_token: ${{ secrets.USER_ACCESS_TOKEN }}


hoangmirs:
Should we consider using [this action](https://github.com/Andrew-Chen-Wang/github-wiki-action)? Then we don't need to care about those values 🤔 
More context: https://nimble-co.slack.com/archives/CSQG2QXFX/p1698199200680339

sanG-github:
Closed this PR since the change is against [this RFC](https://github.com/nimblehq/compass/discussions/1334).

PR #501, opened 2024-01-10
[#429] Revisit the test workflow

### LINKED ISSUE #429: Revisit the test workflow
## Why

Need to double check [this value ](https://github.com/nimblehq/rails-templates/blob/develop/.template/addons/github/.github/workflows/test.yml.tt#L46C71-L46C71)should be `develop` or `development`

image

## Who Benefits?

Developers

sanG-github:
- Close #429

## What happened 👀

Revised the test workflow, at the step using the `branch-tag-action` workflow.

Based on the current implementation of the branch-tag-action workflow, it should be `develop` instead of ~`development`~ since the `BRANCH_TAG` depends on the `branch` or `tag`.

```shell
if [[ $BRANCH_NAME == master || $BRANCH_NAME == main ]]
then
    BRANCH_TAG=latest
elif [[ $TAG_NAME != $REF ]]
then
    BRANCH_TAG=$TAG_NAME
else
    BRANCH_TAG="${BRANCH_NAME//\//-}"
fi
```

## Insight 📝

`n/a`

## Proof Of Work 📹

Small change as described in the above section.

### DIFF
--- a/.template/addons/github/.github/workflows/test.yml.tt
+++ b/.template/addons/github/.github/workflows/test.yml.tt
@@ -45,7 +45,7 @@ jobs:
           password: ${{ env.DOCKER_REGISTRY_TOKEN }}
 
       - name: Pull Docker image
-        if: ${{ env.BRANCH_TAG != 'latest' && env.BRANCH_TAG != 'development' }}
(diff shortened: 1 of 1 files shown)


malparty (approved):

Goose97 (approved):
🚀

PR #504, opened 2024-01-25
[#357] Fix: docker compose command is not working as expected

### LINKED ISSUE #357: Fix: docker compose command is not working as expected
## Issue

- We change from using `docker-compose` command(v1) to `docker compose` command(v2 - rewritten by Go). But it seems it does not work properly. Sometimes it does not recognize environment variables, then it just uses the default value and breaks our workflow.

For example, it sets the wrong image name:
__352__Remove_the_deprecated_desired_capabilities_from_capybara_config_·_nimblehq_rails-templates_c6308ed
 
## Expected

We might need to revert back to using `docker-compose` command. I try reverting in the EWA project and it works perfectly
 
## Steps to reproduce

1. This workflow failed: https://github.com/nimblehq/rails-templates/runs/7803426993?check_suite_focus=true

sanG-github:
- Close #357

## What happened 👀

Update the `docker-compose` from v1 to `docker compose` v2. 🚀 

Update the `APP_NAME` to use `hyphen` instead of `underscore` when it comes to `container_name`.

## Insight 📝

When changing the docker-compose from version 1 to version 2, the gist is that they keep the container name to be a valid name for DNS hostnames with `hyphen (-)` as a separator instead of `underscore (_)` as before. ([ref](https://arc.net/l/quote/imttmosv))

> [!NOTE]  
>  **With the APP_NAME = `rails_template`**

For the current codebase, we previously used `underscore` for all the container names primarily, such as:
  - `rails_template_test_run_9ae09012e3d0`
  - `rails_template_redis`

But after changing docker-compose to a newer version without any additional modifications, those container names become:
  - `rails_template-test-run-9ae09012e3d0`
  - `rails_template-redis`

-> So we need to update all container names to be consistent with that convention, using only `hyphen`.

> [!CAUTION]  
>  We previously tried to find the `test run` container by the name, but it was used with `underscore` -> challenge when switching to use `hyphen` as it keeps raising the container name not found.

I tried to log some information when the test flow was running, but it looked weird.

In this [workflow run](https://github.com/nimblehq/rails-templates/actions/runs/7666956273/job/20895773724), the log returns a list of the container name and ID respectively. And we can get the `ID` from the `container name` as show below:

image

But after changing it to `hyphen`, with the same approach, we cannot retrieve that container ID in this [workflow run](https://github.com/nimblehq/rails-templates/actions/runs/8259147422/job/22592682351).

image

## Proof Of Work 📹

| Built and renamed the docker image successfully                                                                                       |
|---------------------------------------------------------------------------------------------------------------------------------------|
| image |


.template/spec/support/serverspec.rb
```diff
@@ -7,7 +7,8 @@ module ServerSpecHelpers
   # Prebuild and run docker image before running the test
   # Because the docker api does not support docker compose
   def self.test_container
-    container_id = `docker ps -qf "name=#{ENV.fetch('APP_NAME')}_test"`
+    container_name = "#{ENV.fetch('APP_NAME')}-test"
```
malparty, line 10:
```suggestion
    container_name = "#{ENV.fetch('CONTAINERIZED_APP_NAME')}-test"
```

malparty (approved):
Just not sure about the last usage of `APP_NAME` 👀

khangbui22 (commented):
LGTM 🚀

khangbui22 (approved):

.template/spec/support/serverspec.rb
sanG-github: As we don't have that environment variable inside the Docker image, therefore it will raise an error.

image

I tried and I think we can keep using the current one.

malparty: Oh, ok ok, well noted, thanks!

PR #506, opened 2024-03-13
[#493][CRUD] Add the form view templates to the generator

### LINKED ISSUE #493: [CRUD] Add the form view templates to the generator
## Why

The default view generator in Rails does not include Bootstrap classes and has a poor layout. To make the generator usable out of the box, the Rails template must ship with custom view templates.

- This story only focuses on the `new`, `create`, `edit` and `update`  views.
- Ensure the template is using the SLIM engine
- Ensure the `show` view is not generated.

## Who Benefits?

Greenfield project teams.

khangbui22:
- Close #493 

## What happened 👀

- [x] Add `new`, `edit` and `_form` view template
- [x] Tests passed 

## Insight 📝

This PR only focuses on the new, edit, and form view templates

## Proof Of Work 📹

https://github.com/nimblehq/rails-templates/assets/36690900/acae140e-1e35-4055-94d3-d63817ac3740

### DIFF
--- a/.template/addons/crud/lib/templates/slim/scaffold/_form.html.slim.tt
+++ b/.template/addons/crud/lib/templates/slim/scaffold/_form.html.slim.tt
@@ -1,0 +1,28 @@
+= form_with(model: <%= model_resource_name %>) do |form|
+  - if <%= singular_table_name %>.errors.any?
+    div style="color: red"
+      h2 = "#{pluralize(<%= singular_table_name %>.errors.count, "error")} prohibited this <%= singular_table_name %> from being saved:"
+      ul
+        - <%= singular_table_name %>.errors.each do |error|
+          li = error.full_message
+
+<% attributes.each do |attribute| -%>
+  .mb-3
+<% if attribute.password_digest? -%>
+    = form.label :password, class: 'form-label'
+    = form.password_field :password, class: 'form-control'
+
+  .mt-3
+    = form.label :password_confirmation, class: 'form-label'
+    = form.password_field :password_confirmation, class: 'form-control'
+<% elsif attribute.attachments? -%>
+    = form.label :<%= attribute.column_name %>, class: 'form-label'
+    = form.<%= attribute.field_type %> :<%= attribute.column_name %>, multiple: true, class: 'form-control'
+<% else -%>
+    = form.label :<%= attribute.column_name %>, class: 'form-label'
+    = form.<%= attribute.field_type %> :<%= attribute.column_name %>, class: 'form-control'
+<% end -%>
+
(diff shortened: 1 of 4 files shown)


.template/addons/crud/lib/templates/slim/scaffold/_form.html.slim.tt
```diff
@@ -0,0 +1,28 @@
...
+  - if <%= singular_table_name %>.errors.any?
+    div style="color: red"
+      h2 = "#{pluralize(<%= singular_table_name %>.errors.count, "error")} prohibited this <%= singular_table_name %> from being saved:"
+      ul
+        - <%= singular_table_name %>.errors.each do |error|
+          li = error.full_message
+
+<% attributes.each do |attribute| -%>
+  .mb-3
+<% if attribute.password_digest? -%>
```
sanG-github, line 11: Should we add an indentation for the `if else` block as it makes us a bit confused with the loop above? 💭

.template/addons/crud/lib/templates/slim/scaffold/_form.html.slim.tt
```diff
@@ -0,0 +1,28 @@
...
+    = form.label :password, class: 'form-label'
+    = form.password_field :password, class: 'form-control'
+
+  .mt-3
```
sanG-github, line 15: In this case, I think it's odd to have the `mt-3` right after the `mb-3` element, what do you think? 🤔

sanG-github (approved):

.template/addons/crud/lib/templates/slim/scaffold/_form.html.slim.tt
khangbui22: Nice catch 👍 It should be `.mb-3` instead. Fixed in https://github.com/nimblehq/rails-templates/pull/506/commits/7f2273c716eae67c317b227f89e738b32fa8d016

.template/addons/crud/lib/templates/slim/scaffold/_form.html.slim.tt
khangbui22: Since we are using Slim template, the indentation of the html block is important. Thus in this case we cannot indent `.mb-3`. If we are to indent the `if else` block it will look like this:

```
  <% attributes.each do |attribute| -%>
  .mb-3
  <% if attribute.password_digest? -%>
 ```

It also looks a bit weird and confusing for me 😅 . Thus I decided any line that I will put them all the beginning of the line.

malparty (approved):
Thanks :pray: 

Sorry for the review delay 🙇

PR #507, opened 2024-03-14
[#422] Change to use `@nimblehq/eslint-config-nimble-core`

### LINKED ISSUE #422: Change to use `@nimblehq/eslint-config-nimble-core`

sanG-github:
- Close #422

## What happened 👀

As we have separated the previous version of `eslint-config-nimble` into many child packages with the main one being `eslint-config-nimble-core`, and we don't need `eslint` rules for `react`, or `typescript`.

Therefore, we change to use the `@nimblehq/eslint-config-nimble-core` package.

## Insight 📝

> [!warning]  
>  Published a new version for the `core` package, but I forgot to remove the `parser` which hasn't been defined in the corresponding `package.json` -> Needs one more update before using it.

```
Error: Failed to load parser '@typescript-eslint/parser' declared in '.eslintrc » @nimblehq/eslint-config-nimble-core » /home/runner/work/rails-templates/rails-templates/rails_templates/node_modules/@nimblehq/eslint-config-nimble-core/lib/rules/import.js': Cannot find module '@typescript-eslint/parser'
```

After updating the version of the core package, there was a warning about the outdated usage of `.eslintignore`, then we switched to using the `config file` instead.

image

## Proof Of Work 📹

image


malparty (approved):
# 🚀 🥳

byhbt (approved):

PR #509, opened 2024-03-29
[#499] Create a Rubocop cop for ordering instructions in class (e.g. Models)

### LINKED ISSUE #499: Create a Rubocop cop for ordering instructions in class (e.g. Models)
## Why

We have a comprehensive set of style guides in the Compass. Sometimes, it hard to keep them all of your head why developing. Turning it into a custom Rubocop rules can lower the cognitive overhead

- Implement a custom Rubocop to validate instructions order for class (e.g. Models)

## Who Benefits?

Developers on standard admin projects.

Goose97:
- Close #499

## What happened 👀

Enforce class template with custom Rubocop rule

https://nimblehq.co/compass/development/code-conventions/ruby/#convention-40

## Insight 📝

The cop is divided into two phases:
1. Expression categorize phase: every expressions in the class body will classify into one of [these categories](https://github.com/nimblehq/rails-templates/pull/509/files#diff-0386f106cc4b680195ae1e3c9125582be517effb3a99d7260d3aa6aae6229eefR9-R23). If none matches, it'll be `:unknown`
2. Validate the order of the class body expressions according to the template

### Project bootstrap

https://github.com/nimblehq/rails-templates/assets/35915460/fbdb2105-a9c5-4ef8-8e76-0b37ad120cfd

### Linter example

https://github.com/nimblehq/rails-templates/assets/35915460/02e996e5-ee7e-4b09-9861-29716144f434

### DIFF
--- a/rubocop/custom_cops/class_template.rb
+++ b/rubocop/custom_cops/class_template.rb
@@ -1,0 +1,109 @@
+# frozen_string_literal: true
+
+require_relative 'class_template/expression_category'
+
+module CustomCops
+  class ClassTemplate < RuboCop::Cop::Base
+    include CustomCops::ExpressionCategory
+
+    EXPRESSION_TYPE_ORDER = {
+      extend: 0,
+      include: 1,
+      constant_assignment: 2,
+      attribute_macro: 3,
+      other_macro: 4,
+      self_class_block: 5,
+      public_class_method: 5,
+      initialization: 6,
+      public_instance_method: 7,
+      alias: 8,
+      alias_method: 8,
+      protected_instance_method: 9,
+      private_instance_method: 10
+    }.freeze
+
+    # Scan class body
+    def on_class(node)
+      expressions = top_level_expressions(node)
+
+      state = { function_visibility: :public, expression_types: [] }
+      expressions.each_with_object(state) { |expression, acc| process_expression(expression, acc) }
+
+      validate_expressions_order(state[:expression_types])
+    end
+
+    private
+
+    def top_level_expressions(class_node)
+      return [] unless class_node.body
+
+      # Multi-expression body
+      return class_node.body.children if class_node.body.type == :begin
+
+      # Single-expression body
+      [class_node.body]
+    end
+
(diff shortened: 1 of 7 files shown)


malparty:
Sorry I was a bit busy. I'll try to review next Monday (I blocked some time in my calendar for it 🙇 )

rubocop/custom_cops/class_template.rb
```diff
@@ -0,0 +1,109 @@
...
+
+    def validate_expressions_order(expressions)
+      expressions = expressions.filter { _1[:category] != :unknown }
+
+      expressions.each_cons(2) do |first, second|
+        next unless EXPRESSION_TYPE_ORDER[first[:category]] > EXPRESSION_TYPE_ORDER[second[:category]]
+
+        add_offense(
+          second[:expression],
+          message: error_message(second[:category], expressions.pluck(:category))
```
malparty, line 66:
```suggestion
          message: error_message(second[:category], expressions.map { |expression| expression[:category] })
```

I tried in the CX project, but it seems that the Rails helpers could not be loaded. More over, in case we want to reuse this cop in a non Rails project, it can be nice to use pure Ruby only :) 



image

malparty (commented):
So it's working well on my side 🥳 

Just one question: do you think we can improve the message so that it points to the line of code (e.g. "...should come before `audited associated_with: :tenant`") 💭 

Note that this suggestion can be done in a future PR, it's not a blocker 👍 

| Message | Code |
|---|---|
|image|image|

Goose97:
> Just one question: do you think we can improve the message so that it points to the line of code (e.g. "...should come before `audited associated_with: :tenant`") 💭

Great suggestion. Added in d8385ebc436f17dbdd45fb0be013b93ac1b52d43

Your example will now display this error instead:

Screen Shot 2024-04-09 at 17 43 38

malparty (approved):
Approved ahead, but we need to resolve the tests 🙇

sanG-github (approved):

PR #511, opened 2024-04-12
[#460] Revisit and set the default value for some environment variables

### LINKED ISSUE #460: Revisit and set the default value for some environment variables
## Why

- The following variables are required when we deploy a Rails application from our template:
```
  MAILER_DEFAULT_HOST: "localhost"
  MAILER_DEFAULT_PORT: "3000"
  MAILER_SENDER: "Test <noreply@nimblehq.co>"
  AVAILABLE_LOCALES: "en"
  DEFAULT_LOCALE: "en"
  FALLBACK_LOCALES: "en"
```

- But they can be set a default value instead of setting them every time we set up the CD
- IMO, we can use the current value for them when we fetch the value in the code:
```
config.i18n.available_locales = ENV.fetch('AVAILABLE_LOCALES', 'en').split(',')
```
- Sometimes, we forget to set them when deploying, and then it fails but it will take time to investigate that issue with some new members.

## Who Benefits?

Developers

sanG-github:
- Close #460

## What happened 👀

Add the default values for some environment variables to reduce the probability missing them while deploying the application, and can be modified when needed. 🚀 

## Insight 📝

`n/a`

## Proof Of Work 📹

Simple files changed. 💭

### DIFF
--- a/.template/spec/base/config/environments/development_spec.rb
+++ b/.template/spec/base/config/environments/development_spec.rb
@@ -25,7 +25,7 @@ describe 'config/environments/development.rb' do
     <<~RUBY
       config.action_mailer.default_url_options = {
         host: ENV.fetch('MAILER_DEFAULT_HOST', 'localhost'),
-        port: ENV.fetch('MAILER_DEFAULT_PORT')
+        port: ENV.fetch('MAILER_DEFAULT_PORT', '3000')
       }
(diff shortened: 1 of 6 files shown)


malparty (approved):
🚀

byhbt (approved):

